PR #6, opened 2016-07-21
Use AWSKMSClientBuilder

austinmoore-:
I decided to make a new `KmsMasterKeyProviderBuilder` class instead of just using the builder directly inside `KmsMasterKeyProvider`. There are two reasons for this:
1. The motivation behind the client builders in the AWS Java SDK was to remove the need to have/use multiple constructors to build an object, and I felt it would be valuable to carry that reasoning over to this library. Using the builder removes the need to search for the correct constructor, and is a bit more expressive.
2. A separate builder class ensures that code previously written using the old constructors will not break with the introduction of the client builders. I noticed that clients created with the builder are immutable, so any calls to the setters in `KmsMasterKeyProvider` that change the state of the KMS client (e.g. `KmsMasterKeyProvider#setRegion`), will throw an exception, and won't change as expected. This way felt safer.

Also, feel free to share thoughts on how the new builder can be tested. I originally added an `equals` method to KmsMasterKeyProvider, and was going to ensure that both the builder and constructor created objects with the same properties, but because `AWSKMSClient` does not have `equals` overriden, this proved to be difficult.

### DIFF
--- a/src/examples/java/com/amazonaws/crypto/examples/BuilderExample.java
+++ b/src/examples/java/com/amazonaws/crypto/examples/BuilderExample.java
@@ -1,0 +1,71 @@
+package com.amazonaws.crypto.examples;
+
+import com.amazonaws.encryptionsdk.AwsCrypto;
+import com.amazonaws.encryptionsdk.CryptoResult;
+import com.amazonaws.encryptionsdk.kms.KmsMasterKey;
+import com.amazonaws.encryptionsdk.kms.KmsMasterKeyProvider;
+import com.amazonaws.encryptionsdk.kms.KmsMasterKeyProviderBuilder;
+
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * <p>
+ * Encrypts and then decrypts a string under a KMS key
+ *
+ * <p>
+ * Arguments:
+ * <ol>
+ * <li>KMS Key Arn
+ * <li>String to encrypt
+ * </ol>
+ */
+public class BuilderExample {
+
+    private static String keyArn;
+    private static String data;
+
+    public static void main(final String[] args) {
+        keyArn = args[0];
+        data = args[1];
+
+        // Instantiate the SDK
+        final AwsCrypto crypto = new AwsCrypto();
+
+        // Set up the KmsMasterKeyProvider using the defaults provided by the KmsMasterKeyProviderBuilder
+        final KmsMasterKeyProvider prov = KmsMasterKeyProviderBuilder.standard()
+                .withKeyId(keyArn)
+                .build();
+
+        // Encrypt the data
+        //
+        // Most encrypted data should have associated encryption context
+        // to protect integrity. Here, we'll just use a placeholder value.
+        //
+        // For more information see:
+        // blogs.aws.amazon.com/security/post/Tx2LZ6WBJJANTNW/How-to-Protect-the-Integrity-of-Your-Encrypted-Data-by-Using-AWS-Key-Management
+        final Map<String, String> context = Collections.singletonMap("Example", "String");
+
+        final String ciphertext = crypto.encryptString(prov, data, context).getResult();
+        System.out.println("Ciphertext: " + ciphertext);
+
+        // Decrypt the data
+        final CryptoResult<String, KmsMasterKey> decryptResult = crypto.decryptString(prov, ciphertext);
+        // We need to check the encryption context (and ideally key) to ensure that
+        // this was the ciphertext we expected
+        if (!decryptResult.getMasterKeyIds().get(0).equals(keyArn)) {
+            throw new IllegalStateException("Wrong key id!");
+        }
+
+        // The SDK may add information to the encryption context, so we check to ensure
+        // that all of our values are present
+        for (final Map.Entry<String, String> e : context.entrySet()) {
+            if (!e.getValue().equals(decryptResult.getEncryptionContext().get(e.getKey()))) {
+                throw new IllegalStateException("Wrong Encryption Context!");
+            }
+        }
+
+        // Now that we know we have the correct data, we can output it.
+        System.out.println("Decrypted: " + decryptResult.getResult());
+    }
+}
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
@@ -131,7 +131,12 @@ public class KmsMasterKeyProvider extends MasterKeyProvider<KmsMasterKey> implem
         kms_ = kms;
         region_ = region;
         regionName_ = region.getName();
-        kms_.setRegion(region);
+        // When kms_ is created by the builder, it is immutable and will throw an exception when setting something
+        try {
+            kms_.setRegion(region);
+        } catch (UnsupportedOperationException e) {
+            // Ignored. Not a problem, region should already be set by the builder.
+        }
         keyIds_ = new ArrayList<>(keyIds);
     }
 
@@ -244,7 +249,7 @@ public class KmsMasterKeyProvider extends MasterKeyProvider<KmsMasterKey> implem
         return region_;
     }
 
-    private static Region getStartingRegion(final String keyArn) {
+    static Region getStartingRegion(final String keyArn) {
         final String region = parseRegionfromKeyArn(keyArn);
         if (region != null) {
             return Region.getRegion(Regions.fromName(region));
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
@@ -140,6 +140,10 @@ public class KmsMasterKeyProvider extends MasterKeyProvider<KmsMasterKey> implem
         keyIds_ = new ArrayList<>(keyIds);
     }
 
+    public static KmsMasterKeyProviderBuilder builder() {
+        return KmsMasterKeyProviderBuilder.standard();
+    }
+
     /**
      * Returns "aws-kms"
      */
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
@@ -220,6 +220,8 @@ public class KmsMasterKeyProvider extends MasterKeyProvider<KmsMasterKey> implem
      * Configures this provider to use a custom endpoint. Sets the underlying {@link Region} object
      * to {@code null}, and instructs the internal KMS client to use the specified {@code endPoint}
      * and {@code regionName}.
+     *
+     * Note: This method will not work if this object was created by {@link KmsMasterKeyProviderBuilder}.
      */
     public void setCustomEndpoint(final String regionName, final String endPoint) {
         if (kms_ instanceof AWSKMSClient) {
@@ -235,6 +237,8 @@ public class KmsMasterKeyProvider extends MasterKeyProvider<KmsMasterKey> implem
     /**
      * Set the AWS region of the AWS KMS service for access to the master key. This method simply
      * calls the same method of the underlying {@link AWSKMSClient}
+     *
+     * Note: This method will not work if this object was created by {@link KmsMasterKeyProviderBuilder}.
      *
      * @param region
      *            string containing the region.
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
@@ -1,7 +1,10 @@
 package com.amazonaws.encryptionsdk.kms;
 
-import com.amazonaws.ClientConfiguration;
-import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.ClientConfigurationFactory;
+import com.amazonaws.client.AwsSyncClientParams;
+import com.amazonaws.client.builder.AwsClientBuilder;
+import com.amazonaws.regions.AwsRegionProvider;
+import com.amazonaws.regions.DefaultAwsRegionProviderChain;
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.RegionUtils;
 import com.amazonaws.regions.Regions;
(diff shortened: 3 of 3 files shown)


src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+ * </pre>
+ *
+ * Example, specifying a region:
+ * <pre>
+ *     KmsMasterKeyProvider keyProvider = KmsMasterKeyProviderBuilder.standard()
+ *       .withRegion("us-east-1")
+ *       .build();
+ * </pre>
+ */
+public class KmsMasterKeyProviderBuilder {
```
SalusaSecondus: This is already missing withMetricsCollector() and withRequestHandlers(). Does it make sense to simply inherit from AwsClientBuilder so that you always get the appropriate (and new) configuration as available?  That said, I'm worried that you might need to use reflection (possibly Proxies, which would be ugly) to properly plumb all current and future calls to your inner client builder.

At the very least, please add withMetricsCollector() and withRequestHandlers() and consider more future proof solutions.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
```diff
@@ -13,11 +16,10 @@
...
 
 /**
- * Used to build a {@link KmsMasterKeyProvider} using {@link AWSKMSClientBuilder}, which will use provider chains to get
- * defaults if the properties are not explicitly set.
+ * Used to build a {@link KmsMasterKeyProvider} which will use the defaults found in the relevant provider chains.
  *
- * Default usage will use all the defaults found by AWSKMSClientBuilder. If a region is not found in the configuration
- * chain, then {@link Regions#DEFAULT_REGION} is used.
+ * If a region is not defined and no region can be found in the region provider chain, then
+ * {@link Regions#DEFAULT_REGION} is used.
```
austinmoore-: I am using `Regions#DEFAULT_REGION` if no region can be found, because that is what the constructors are currently doing. However, the Java SDK now throws an exception if a region cannot be found in the region provider chain. Should the builder follow the Java SDK builders? More on why: 

> If you create a client through the constructor, it initializes the client with a default endpoint (typically, us-east-1 or a global endpoint). For this reason, many customers don't bother setting a region because it just works with the default endpoint. When these customers attempt to deploy their application to another region, a lot of code has to be changed to configure the client with the correct region based on the environment.

https://java.awsblog.com/post/Tx3QE4T4596T5G4/Fluent-Client-Builders

pom.xml
```diff
@@ -37,7 +37,7 @@
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk</artifactId>
-      <version>1.10.61</version>
+      <version>1.11.18</version>
```
pravsingh: Upgrade & Builder change both of the part of the PR. Shouldn't the title say so?

src/examples/java/com/amazonaws/crypto/examples/BuilderExample.java
```diff
@@ -0,0 +1,71 @@
+package com.amazonaws.crypto.examples;
```
pravsingh: Examples are great however we should prefer unit test cases as they assert the functionality as part of the build and also show the usage of it.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -125,7 +125,12 @@ protected KmsMasterKeyProvider(final AWSKMSClient kms, final Region region, fina
         kms_ = kms;
         region_ = region;
         regionName_ = region.getName();
-        kms_.setRegion(region);
+        // When kms_ is created by the builder, it is immutable and will throw an exception when setting something
+        try {
+            kms_.setRegion(region);
+        } catch (UnsupportedOperationException e) {
+            // Ignored. Not a problem, region should already be set by the builder.
```
pravsingh: 👍

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+    private Region determineRegion() {
+        Region region = RegionUtils.getRegion(this.getRegion());
+
+        final AwsRegionProvider regionProvider = new DefaultAwsRegionProviderChain();
+
+        if (region != null) {
+            return region;
+        } else {
+            final String regionName = regionProvider.getRegion();
+            if (regionName != null) {
```
pravsingh: instead of using US_WEST_2 as default, could we use Regions.getCurrentRegion()?

src/test/java/com/amazonaws/services/kms/MockKMSClient.java
```diff
@@ -83,7 +93,7 @@
     private Region region_ = Region.getRegion(Regions.DEFAULT_REGION);
 
     @Override
-    public void createAlias(CreateAliasRequest arg0) throws AmazonServiceException, AmazonClientException {
+    public CreateAliasResult createAlias(CreateAliasRequest arg0) throws AmazonServiceException, AmazonClientException {
```
pravsingh: I see changes in the contract here.

doesn't it break the existing test cases written by developers in their own applications?

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
SalusaSecondus: Where do you see us using us-west-2 as default?

src/test/java/com/amazonaws/services/kms/MockKMSClient.java
bdonlan: This class is not part of the public API of the SDK. Any changes here are adapting to upstream AWS SDK changes.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+ * </pre>
+ *
+ * Example, specifying a region:
+ * <pre>
+ *     KmsMasterKeyProvider keyProvider = KmsMasterKeyProviderBuilder.standard()
+ *       .withRegion("us-east-1")
+ *       .build();
+ * </pre>
+ */
+public class KmsMasterKeyProviderBuilder extends AwsClientBuilder<KmsMasterKeyProviderBuilder, KmsMasterKeyProvider> {
```
bdonlan: So, one thing I note is that the Python SDK supports using key IDs from multiple regions in a single KMS MKP. I'd prefer to align the Java behavior more closely with the Python behavior in this regard, and we need to make sure the builder API we're creating here is amenable to this.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private KmsMasterKeyProviderBuilder() {
+        super(new ClientConfigurationFactory());
+    }
+
+    /**
+     * Sets the region to be used by the client. Overrides any previously set region.
+     */
+    public KmsMasterKeyProviderBuilder withRegion(Region region) {
```
bdonlan: I'd prefer to call this `withDefaultRegion` to leave open the possibility of supporting keys from multiple regions in the same client in future versions.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+        return withRegion(region.getName());
+    }
+
+    /**
+     * Sets the region using a keyId. This will override any previously set regions.
```
bdonlan: We should only infer the region from a key ID if the user hasn't manually configured a region, to avoid confusion. The rule is that we'd use `withDefaultRegion` first, followed by the implied region in `withKeyId`, followed by the SDK default region.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -125,7 +125,12 @@ protected KmsMasterKeyProvider(final AWSKMSClient kms, final Region region, fina
         kms_ = kms;
```
bdonlan: We should `@Deprecated` the old constructors and encourage folks to move to the builder API.

I'd also suggest a `builder()` static method on `KmsMasterKeyProvider` for consistency with other builders in the SDK.

src/examples/java/com/amazonaws/crypto/examples/BuilderExample.java
bdonlan: Yes, and we should also just update existing examples to use the new API instead of adding a new bespoke example.

pom.xml
```diff
@@ -37,7 +37,7 @@
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk</artifactId>
```
bdonlan: This will probably need a rebase.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
bdonlan: This looks like it's only a fallback in the case where the `DefaultAwsRegionProviderChain` fails. Which is reasonable enough,

src/test/java/com/amazonaws/services/kms/MockKMSClient.java
bdonlan: This is not part of the public API as it's in the test directory. It's also responding to breaking SDK API changes (i.e. it's a build fix after the dependency update).

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
bdonlan: This looks like a fallback in the case where the SDK's default region selection logic fails, which is reasonable enough.

bdonlan:
I should note that it would be useful to be able to reuse KMS clients, but I don't want to tie us to an API for that until we know what multi-region KMS MKPs would look like.

bdonlan:
Regarding testing, we'll need to figure out how we want to run automated integration tests for this. We have some cross-language compatibility tests that operate on KMS, but I'm not sure if we've put those in a public repository just yet, and we certainly haven't set up a publicly visible test fleet using actual AWS credentials to run KMS calls. When we do that's probably the best place to put tests for client construction.

src/examples/java/com/amazonaws/crypto/examples/BuilderExample.java
pravsingh: @bdonlan,
 Since you are the contributor and know more about this project than I do.
I have commented it in other PR and asking the same question here again.
Do you think we should put examples as part of the main sdk library,
or we should either create separate maven modules or keep them under src/test?
Usually the tendency is that being it an example, developers can put less secured code and insensitive logging or other kinds of code which could be prone to security threats. e.g. example providing public methods or utils which could reveal secrets. This is the one example I can think of right now however there could be more.
just a thought.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
pravsingh: @SalusaSecondus,
in Regions.java inside aws-java-sdk-core:
```java
 /**
     * The default region that new customers in the US are encouraged to use
     * when using AWS services for the first time.
     */
    public static final Regions DEFAULT_REGION = US_WEST_2;
```

@bdonlan agree with you. looks good to me.

src/examples/java/com/amazonaws/crypto/examples/BuilderExample.java
bdonlan: Generally I think if we're checking in example code we should make sure it's _production quality_ example code that we want people to imitate. We should also avoid pulling in new dependencies for examples in the main repository. For the more complex example you refer to, since it involves a lot of moving parts (a client component, a server in lambda, a cloudformation template...) it might be preferable to split it off to avoid confusion - or maybe have it as a separate maven subproject in a subdirectory? I'm not enough of a maven expert to know if that is a good approach here though.

pravsingh: agree with you.
I have maven expertise and will create PR so that you can review if it makes sense.

pravsingh: created https://github.com/awslabs/aws-encryption-sdk-java/issues/32

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
austinmoore-: I believe this is supported with [`withKeyIds`](https://github.com/awslabs/aws-encryption-sdk-java/pull/6/files#diff-854ddaccf0f12318e57edc82ddd8ea44R77).

`withKeyIds` does not set a region, so you could use `withDefaultRegion`, add multiple key IDs, and it should create a KMS MKP for the region you set in `withDefaultRegion`.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProviderBuilder.java
austinmoore-: This sounds good to me. Should we throw an exception if the user tries to use `withKeyId` after they have already set a default region? Or just continue with the region set with `withDefaultRegion`?

I'm leaning towards being explicit, and throwing an exception so that there is no unexpected behavior on the user's end. But I could be swayed.

bdonlan:
Thanks for the submission, and sorry for the radio silence - it wasn't obvious that progress was happening on this for a while and it dropped off my raday. In the meantime, I've had some work-in-progress code that did something like this kicking around for a while in my local git repo and finally put it up as PR #40. The main notable differences are that the one I submitted supports contacting multiple regions from the same MKP, and marks a bunch of older methods as now being deprecated. It also includes tests ;) I'd encourage you to check it out and see if it is along the lines of what you wanted to see from the API.

bdonlan:
Closing this as moot at this point as we merged a different version of the builder concept.

PR #7, opened 2016-07-26
Support for EC without point compression

austinmoore-:
This PR makes using point compression optional. Disabling point compression is done like so:

``` java
final AwsCrypto crypto = new AwsCrypto();
crypto.setPointCompressionEnabled(false);
```

Decryption is unaffected by point compression because BC supports both.

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
@@ -87,6 +87,8 @@ import com.amazonaws.util.Base64;
 public class AwsCrypto {
     private static final Map<String, String> EMPTY_MAP = Collections.emptyMap();
 
+    private boolean pointCompressionEnabled_ = true;
+
     /**
      * Returns the {@link CryptoAlgorithm} to be used for encryption when none is explicitly
      * selected. Currently it is {@link CryptoAlgorithm#ALG_AES_256_GCM_IV12_TAG16_HKDF_SHA384_ECDSA_P384}.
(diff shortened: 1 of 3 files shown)


src/test/java/com/amazonawsencryptionsdk/AwsCryptoTest.java
```diff
@@ -159,6 +159,35 @@ public void encryptDecrypt() {
     }
 
     @Test
+    public void encryptDecryptPointCompressionDisabled() {
```
pravsingh: thanks for the test case. 👍

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -87,6 +87,8 @@
 public class AwsCrypto {
     private static final Map<String, String> EMPTY_MAP = Collections.emptyMap();
 
+    private boolean pointCompressionEnabled_ = true;
```
pravsingh: this keeps it backward compatible.

could you write a test case to ensure that the default value is always intact and is set to true?

SalusaSecondus:
I actually forgot about this PR. What is the value of this change?

bdonlan:
Currently, the [format spec](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/message-format.html#header-aad) specifies using compressed points, and in fact both our Java and Python SDKs make use of compressed points on encrypt.

One of the goals of the encryption SDKs is to avoid format fragmentation; we want to ensure that if any compliant SDK encrypts a message, then as long as there's at least one Master Key that is usable on the decrypt side, the message can be decrypted. This means the decrypt side must always support a superset of whatever the encrypt side supports, for core format options. Introducing this as an _option_ on the encrypt side expands the set of behavior that decrypt-side implementations are _required_ to support.

I'm not sure if we accidentally support uncompressed points on decrypt currently; if we do, this is unintentional behavior and subject to change in the future; we may even want to explicitly remove support to reduce the risk that someone will take a dependency on this behavior.

I'll also note that the only time this support is useful is when, on the decrypt side, point compression is not yet implemented in platform or language libraries. In this case, it's already impossible to decrypt messages encrypted with the current released version of the library, and we thus already have format fragmentation. We'd be inclined to implement point (de)compression on the decrypt side - thus fully resolving the fragmentation issue - instead of maintaining what is essentially an incompatible, parallel data format.

For these reasons I'm afraid we can't accept this contribution; if you really want to use uncompressed points for some reason, you could write a custom `CryptoMaterialsManager` that uses an uncompressed point format, without needing to modify the core SDK (this is a new extension point added after this PR was originally opened).

PR #12, opened 2016-11-09
Address bug in encryptionContext check.

conor10:
The check should be dictated by the encryptionContext, not the provided context, as in it's existing form, if an empty context is provided, the decrypt operation would still be successful. This would bypass the point of the encryptionContext entirely.

### DIFF
--- a/src/examples/java/com/amazonaws/crypto/examples/StringExample.java
+++ b/src/examples/java/com/amazonaws/crypto/examples/StringExample.java
@@ -68,8 +68,8 @@ public class StringExample {
 
         // The SDK may add information to the encryption context, so we check to ensure
         // that all of our values are present
-        for (final Map.Entry<String, String> e : context.entrySet()) {
-            if (!e.getValue().equals(decryptResult.getEncryptionContext().get(e.getKey()))) {
+        for (final Map.Entry<String, String> e : decryptResult.getEncryptionContext().entrySet()) {
+            if (!e.getValue().equals(context.get(e.getKey()))) {
                 throw new IllegalStateException("Wrong Encryption Context!");
             }
         }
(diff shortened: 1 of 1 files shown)


src/examples/java/com/amazonaws/crypto/examples/StringExample.java
```diff
@@ -68,8 +68,8 @@ public static void main(final String[] args) {
 
         // The SDK may add information to the encryption context, so we check to ensure
         // that all of our values are present
-        for (final Map.Entry<String, String> e : context.entrySet()) {
-            if (!e.getValue().equals(decryptResult.getEncryptionContext().get(e.getKey()))) {
+        for (final Map.Entry<String, String> e : decryptResult.getEncryptionContext().entrySet()) {
+            if (!e.getValue().equals(context.get(e.getKey())) && !e.getKey().equals(Constants.EC_PUBLIC_KEY_FIELD)) {
```
pravsingh: could you please write the test case? Also whether this change is going to be backward compatible.

SalusaSecondus:
Dictating it by the encryption context in the ciphertext means that the data being checked is entirely under the control of a potential attacker (and thus less useful). Generally, when checking AAD or similar structures (such as Encryption Context) you use a locally generated "known good" value to check against.

The check as we wrote it means that you'll miss any *additional* attributes, but that you'll check all that you are aware of (which in the simple case are all that you care about).  In the other direction, they could simply omit attributes that they don't want you to check.

In the highest security cases, you ensure that every value you care about (both present and absent) is correct. This, however, is application specific and thus challenging to show in an example.

Back to the main concern though. You are correct that if a developer passes in an empty encryption context to this method that all encryption contexts will be accepted. However, we can generally assume that it is harder for an attacker to change the logical flow of the program than to provide a ciphertext with different (and unexpected) encryption context. So, this is the safer of the two options.

bdonlan:
We could consider masking SDK-internal context keys from caller code, but that's also a breaking change.

Since we want to reserve the right to add additional SDK internal context keys in the future, I'm inclined to keep the current logic.

PR #26, opened 2017-09-13
Add data key caching examples

juneb:


src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/SimpleDataKeyCachingExample.java
```diff
@@ -0,0 +1,79 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import javax.xml.bind.DatatypeConverter;
```
bdonlan: Stray import? This one will cause issues on Java 9 in particular.

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/cloudformation-template.yaml
```diff
@@ -0,0 +1,221 @@
+Parameters:
```
bdonlan: burying this file under a deep path where people are probably only expecting java source is probably less than ideal

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/SimpleDataKeyCachingExample.java
```diff
@@ -0,0 +1,79 @@
...
+        byte[] myData = "My plaintext data".getBytes(StandardCharsets.UTF_8);
+
+        // Encryption context
+        final Map<String, String> encryptionContext = Collections.singletonMap("purpose", "test");
+
+        // Create a master key provider
+        MasterKeyProvider<KmsMasterKey> keyProvider = new KmsMasterKeyProvider(kmsCmkArn);
+
+        // Create a cache
+        CryptoMaterialsCache cache = new LocalCryptoMaterialsCache(cacheCapacity);
```
bdonlan: This cache is not very helpful, since it's only being used for a single request. This needs to be refactored to keep the caching CMM around across multiple requests.

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/LambdaDecryptAndWrite.java
```diff
@@ -0,0 +1,89 @@
...
+import com.amazonaws.services.lambda.runtime.events.KinesisEvent.KinesisEventRecord;
+import com.amazonaws.util.BinaryUtils;
+
+/**
+ * Decrypts all incoming Kinesis records and writes records to DynamoDB.
+ */
+public class LambdaDecryptAndWrite {
+    private static final long MAX_ENTRY_AGE_MILLISECONDS = 600000;
+    private static final int MAX_CACHE_ENTRIES = 100;
+    private CachingCryptoMaterialsManager cachingMaterialsManager_;
```
pravsingh: what kind of naming convention it is which ends with '_'?
never heard of such anywhere else. May be I'm not aware of the convention in this repo but curious to know.
thanks

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/LambdaDecryptAndWrite.java
```diff
@@ -0,0 +1,89 @@
...
+    private AwsCrypto crypto_;
+    private Table table_;
+
+    /**
+     * Because the cache is used only for decryption, the code doesn't set
+     * the max bytes or max message security thresholds that are are enforced
+     * only on on data keys used for encryption.  
+     */
+    public LambdaDecryptAndWrite() {
+    	String cmkArn = System.getenv("CMK_ARN");
```
pravsingh: do we have a centralized class defining constants? e.g. CMK_ARN

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/LambdaDecryptAndWrite.java
```diff
@@ -0,0 +1,89 @@
...
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except
+ * in compliance with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.amazonaws.crypto.examples.kinesisdatakeycaching;
```
pravsingh: should examples (anything inside com.amazonaws.crypto.examples package) go in src/test or totally new maven module?
Not a good idea to keep example code in final sdk code. It can be separate for references purposes.
This gives flexibility on users choosing what to add as a dependency.

being 'examples', it gives future developers an idea of keeping test codes, utils or some insecured code or doing unwanted logging which could reveal more than we expect.

thoughts?

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/LambdaDecryptAndWrite.java
```diff
@@ -0,0 +1,89 @@
...
+import com.amazonaws.encryptionsdk.caching.CachingCryptoMaterialsManager;
+import com.amazonaws.encryptionsdk.caching.LocalCryptoMaterialsCache;
+import com.amazonaws.encryptionsdk.kms.KmsMasterKey;
+import com.amazonaws.encryptionsdk.kms.KmsMasterKeyProvider;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder;
+import com.amazonaws.services.dynamodbv2.document.DynamoDB;
+import com.amazonaws.services.dynamodbv2.document.Item;
+import com.amazonaws.services.dynamodbv2.document.Table;
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.events.KinesisEvent;
```
pravsingh: why are we gluing it with a specific caching solution e.g. Kinesis and why can't we have provider mechanism so that users can choose the caching solution they use?

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/cloudformation-template.yaml
```diff
@@ -0,0 +1,221 @@
...
+                                Resource: !Sub arn:aws:kinesis:${AWS::Region}:${AWS::AccountId}:stream/${InputStream}
+    JavaLambdaFunction:
+        Type: AWS::Lambda::Function
+        Properties:
+            Description: Java consumer
+            Runtime: java8
+            MemorySize: 512
+            Timeout: 90
+            Role: !GetAtt JavaLambdaRole.Arn
+            Handler: com.amazonaws.crypto.examples.kinesisdatakeycaching.LambdaDecryptAndWrite::handleRequest
```
pravsingh: Kinesis with Lambda should be separate maven module.
Shouldn't add dependency on Kinesis to existing applications which don't use these services.

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/LambdaDecryptAndWrite.java
bdonlan: This is a good point. We should probably move this stuff into a different repository and just reference it from README.md.

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/LambdaDecryptAndWrite.java
bdonlan: It's the convention used in a lot of the code in the repo. It's debatable whether it should be used (I don't much care for it personally) but until that debate happens and the existing code reformatted, consistency with existing code is generally a good thing.

src/examples/java/com/amazonaws/crypto/examples/DataKeyCaching/cloudformation-template.yaml
bdonlan: Note that this is the example from the docs: http://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/sample-cache-example-java.html

Kinesis is shown as a motivating example of where you might want to use the caching feature.

WesleyRosenblum:
Data key caching examples have been added to the keyring branch.

Issue #29, opened 2017-09-15
estimateCiphertextSize is not cached


bdonlan:
estimateCiphertextSize doesn't pass plaintext size down to the CMM, so the cache is always bypassed.

PR #30, opened 2017-09-16
Make estimateCiphertextSize work with cached keys

bdonlan:
estimateCiphertextSize was not passing a plaintext size down to the CMM, so the
caching CMM assumed that a streaming encryption of unknown size was being
performed, and bypassed the cache entirely.

This change passes a plaintext size of zero instead to allow cached keys to be
used; since we don't actually encrypt any data it's safe to not consume any of
the byte limit.

Note that this may not behave quite right if the CMM does more clever things
with plaintext size. In general we should probably resolve this by moving
estimateCiphertextSize over to the Cipher*Stream objects instead, where it can
know the actual DataKeys in use instead of hoping they're reasonably
consistently sized.

Fixes: #29


mattsb42-aws:
LGTM. I also checked the Python client to verify, and this issue is not present there due to the way that ciphertext size estimation is implemented.

https://github.com/awslabs/aws-encryption-sdk-python/blob/master/src/aws_encryption_sdk/streaming_client.py#L375

Issue #32, opened 2017-09-19
refactor examples into its own module


pravsingh:
per conversation at https://github.com/awslabs/aws-encryption-sdk-java/pull/6#discussion_r139338345, creating this issue.

```bash
aws-encryption-sdk-java\
->    pom.xml   //artifact name for this can be aws-encryption-sdk-java-parent
->    aws-encryption-sdk-java-core
        -> pom.xml, src   //this generates aws-encryption-sdk-java artifact
->    aws-encryption-sdk-java-examples
       -> pom.xml, src  //this generates aws-encryption-sdk-java-examples artifact
```

Guys put your feedbacks/proposals in the comments section.

PR #33, opened 2017-09-20
separate module for examples

pravsingh:
Ref: https://github.com/awslabs/aws-encryption-sdk-java/issues/32

Please review. I'll incorporate your feedbacks in my next commit.

FYI @bdonlan


bdonlan:
Unfortunately it looks like this branch currently conflicts and can't be merged. There are also some other changes in e.g. PR #40 that will touch those same files soon, so it's best not to rebase just yet.

That being said, I'm thinking the better approach here would be to move them off to a separate repo - I don't want to have the version number in multiple places and add more levels of directories to the source tree if we can avoid it. I'll look into getting a new repo set up on our side.

bdonlan:
Closing this PR for now, we'll track the progress in #32 until we can finish setting up the new examples repo.

PR #34, opened 2017-09-20
Switch to supported ClientConfiguration.withUserAgentSuffix

SalusaSecondus:
The previously method to append user-agent information (`request.getRequestClientOptions().appendUserAgent()`) really is an internal method and we should avoid using it if possible. Now that `ClientConfiguration` supports `withUserAgentSuffix()`, we're switching to it.

Modification to .gitignore is to help it work better in my workspace.

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
@@ -32,6 +32,7 @@ import com.amazonaws.encryptionsdk.MasterKeyRequest;
 import com.amazonaws.encryptionsdk.exception.AwsCryptoException;
 import com.amazonaws.encryptionsdk.exception.NoSuchMasterKeyException;
 import com.amazonaws.encryptionsdk.exception.UnsupportedProviderException;
+import com.amazonaws.encryptionsdk.internal.VersionInfo;
 import com.amazonaws.internal.StaticCredentialsProvider;
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.Regions;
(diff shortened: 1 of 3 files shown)


mattsb42-aws:
Looks good to me, but I'll hold off on a more experienced Java eye before pulling.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -110,7 +113,10 @@ public KmsMasterKeyProvider(final AWSCredentialsProvider creds) {
      */
     public KmsMasterKeyProvider(final AWSCredentialsProvider creds, final Region region,
             final ClientConfiguration clientConfiguration, final String keyId) {
-        this(new AWSKMSClient(creds, clientConfiguration), region, Collections.singletonList(keyId));
+        this(
+            new AWSKMSClient(creds, clientConfiguration.withUserAgentSuffix(VersionInfo.USER_AGENT)),
```
bdonlan: Doesn't this modify the passed-in ClientConfiguration?

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -119,7 +125,10 @@ public KmsMasterKeyProvider(final AWSCredentialsProvider creds, final Region reg
      */
     public KmsMasterKeyProvider(final AWSCredentialsProvider creds, final Region region,
             final ClientConfiguration clientConfiguration, final List<String> keyIds) {
-        this(new AWSKMSClient(creds, clientConfiguration), region, keyIds);
+        this(
+            new AWSKMSClient(creds, clientConfiguration.withUserAgentSuffix(VersionInfo.USER_AGENT)),
```
bdonlan: Same here

SalusaSecondus: I'll fix this.

PR #40, opened 2017-11-08
Add a builder API for constructing KMS MKPs

bdonlan:
The current KMS MKP constructors use legacy (@Deprecated) KMS client
constructors, and are likely to break at some point in the future. In addition,
they also mutate the client that is passed in, and have an unfortunate
combinatorial explosion of argument types.

This change introduces a new builder API for constructing KMS MKPs, and
deprecates the old one. In addition, it adds support for decrypting KMS keys
from multiple regions with the same MKP, bringing us to feature parity with the
Python SDK.

For now, the semantics of code using the old constructors is unchanged, but
it's likely that we'll want to remove these constructors the next time we make
a breaking change to our APIs.


bdonlan:
Note that this change preserves the behavior that `new KmsMasterKeyProvider()` binds to the `us-west-2` region, ignoring environment variable/system property/etc default region selection. This is why the default constructor is marked deprecated in this change.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -270,6 +537,20 @@ private static Region getStartingRegion(final String keyArn) {
...
 
+    private String identifyKeyRegion(final String keyArn) {
+        String region = parseRegionfromKeyArn(keyArn);
+
+        if (region != null) {
+            return region;
+        }
+
+        if (defaultRegion_ == null) {
+            throw new AwsCryptoException("Can't use non-ARN key identifiers or aliases when no default region is set");
```
bdonlan: We should probably complain on `build()` actually - since the list of key IDs for encrypt can't change after construction using the builder, and this isn't relevant on decrypt.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -32,76 +38,348 @@
...
+            return this;
+        }
+
+        /**
+         * Adds grant tokens to the provider under construction.
+         *
+         * @param tokens
+         * @return
+         */
+        public Builder withGrantTokens(List<String> tokens) {
```
SalusaSecondus: GrantTokens are fundamentally regional and should be treated that way. There is no value (and some negative) to passing them into the wrong region.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -208,53 +484,44 @@ public KmsMasterKey getMasterKey(final String provider, final String keyId) thro
...
     }
 
+    /**
+     * @deprecated This method is inherently not thread safe. Use {@link Builder#setGrantTokens(List)} instead.
+     * {@link KmsMasterKeyProvider}s constructed using the builder will throw an exception on attempts to modify the
+     * list of grant tokens.
+     */
+    @Deprecated
     @Override
     public void setGrantTokens(final List<String> grantTokens) {
```
SalusaSecondus: As noted elsewhere, GrantTokens are regional things.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -32,76 +38,348 @@
...
+         * @return The client to use, or null if this region cannot or should not be used.
+         */
+        AWSKMS getClient(String regionName);
+    }
+
+    public static class Builder implements Cloneable {
+        private String defaultRegion_ = null;
+        private RegionalClientSupplier regionalClientSupplier_ = null;
+        private AWSKMSClientBuilder templateBuilder_ = null;
+        private List<String> keyIds_ = new ArrayList<>();
```
SalusaSecondus: This should be copied when .clone() is invoked. A shallow copy here will likely cause problems.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -32,76 +38,348 @@
...
+         * @param regionName The region to get a client for
+         * @return The client to use, or null if this region cannot or should not be used.
+         */
+        AWSKMS getClient(String regionName);
+    }
+
+    public static class Builder implements Cloneable {
+        private String defaultRegion_ = null;
+        private RegionalClientSupplier regionalClientSupplier_ = null;
+        private AWSKMSClientBuilder templateBuilder_ = null;
```
SalusaSecondus: This should be copied when .clone() is invoked. A shallow copy here will likely cause problems.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -32,76 +38,348 @@
...
+        private List<String> keyIds_ = new ArrayList<>();
+        private List<String> grantTokens_ = new ArrayList<>();
```
SalusaSecondus: This should be copied when .clone() is invoked. A shallow copy here will likely cause problems.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
mattsb42-aws: ++

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKeyProvider.java
```diff
@@ -70,12 +70,27 @@
...
+    public static final class Builder implements Cloneable {
         private String defaultRegion_ = null;
         private RegionalClientSupplier regionalClientSupplier_ = null;
         private AWSKMSClientBuilder templateBuilder_ = null;
         private List<String> keyIds_ = new ArrayList<>();
-        private List<String> grantTokens_ = new ArrayList<>();
+
+        public Builder clone() {
+            try {
+                Builder cloned = (Builder)super.clone();
```
SalusaSecondus: NitPick: space after the cast

PR #44, opened 2017-12-05
Restore logic to set UA suffixes

bdonlan:
The logic to set the UA suffix to include the SDK version was dropped as part of the new builder refactoring. This change restores it.

Also included is a minor fix to a test (this didn't break anything but should be fixed out of general tidiness).

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKey.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKey.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 
 import com.amazonaws.AmazonServiceException;
+import com.amazonaws.AmazonWebServiceRequest;
 import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.encryptionsdk.AwsCrypto;
@@ -33,6 +34,7 @@ import com.amazonaws.encryptionsdk.MasterKey;
 import com.amazonaws.encryptionsdk.MasterKeyProvider;
 import com.amazonaws.encryptionsdk.exception.AwsCryptoException;
 import com.amazonaws.encryptionsdk.exception.UnsupportedProviderException;
+import com.amazonaws.encryptionsdk.internal.VersionInfo;
 import com.amazonaws.services.kms.AWSKMS;
 import com.amazonaws.services.kms.model.DecryptRequest;
 import com.amazonaws.services.kms.model.DecryptResult;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKey.java
```diff
@@ -51,6 +57,75 @@
...
+
+    private final String userAgent_;
+
+    static {
+        Field clientConfiguration = null;
+        try {
+            clientConfiguration = AmazonWebServiceClient.class.getDeclaredField("clientConfiguration");
+            clientConfiguration.setAccessible(true);
+        } catch (Throwable t) {
+            // Incompatible SDK change, we'll leave the Field null. Later we'll just replace the entire UA.
```
SalusaSecondus: Please set `clientConfiguration` to `null` in this catch block.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKey.java
```diff
@@ -51,6 +57,75 @@
...
+        }
+
+        f_clientConfiguration = clientConfiguration;
+    }
+
+    private static String getOriginalUserAgent(AWSKMS kms) {
+        try {
+            ClientConfiguration clientConfiguration = null;
+            if (kms instanceof AWSKMSClient) {
+                clientConfiguration = (ClientConfiguration)f_clientConfiguration.get(kms);
```
SalusaSecondus: What if `f_clientConfiguration == null`?

NitPick: I know we don't do it everywhere, but I much prefer a space after a cast.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKey.java
```diff
@@ -51,6 +57,75 @@
...
+
+                if (suffix == null) {
+                    suffix = "";
+                }
+
+                return prefix + suffix;
+            }
+        } catch (Throwable t) {
+            // fall through
+            // note: Throwable catch is required because we are calling invokeExact
```
SalusaSecondus: This is no longer accurate. Please fix the comment and the catch-block.

src/main/java/com/amazonaws/encryptionsdk/kms/KmsMasterKey.java
```diff
@@ -51,6 +57,75 @@
...
+
+    private static String getUserAgent(AWSKMS kms) {
+        String ua;
+        String baseUserAgent = getOriginalUserAgent(kms);
+
+        if (baseUserAgent == null) {
+            // The internal SDK fields we use to extract the base user agent are missing. We'll override the UA entirely.
+            ua = VersionInfo.USER_AGENT;
+        } else if (baseUserAgent.isEmpty()) {
+            // We need to ensure we don't start with an whitespace character as this breaks client-side signing logic.
```
SalusaSecondus: Where do we ensure this? Not here, so the comment here isn't the right place.

bdonlan:
After discussion with the Java SDK folks, I found a simpler way to update the UA. The appendUserAgent API is documented as internal but they say we should be fine to use it.

PR #59, opened 2018-07-20
Log a warning when an unsupported asym algorithm is used with JceMasterKey

### LINKED ISSUE #56: Define constraints on JceMasterKey RSA wrapping algorithms
## Problem

`JceMasterKey` is very poorly constrained when using RSA master keys. [The current implementation][1] only verifies that the wrapping algorithm name starts with `"RSA/ECB/"`, leaving the padding algorithm open to whatever any given JCE provider makes available.

https://github.com/awslabs/aws-encryption-sdk-python/issues/56

Without providing some constraints around the permitted padding algorithms for RSA master keys, we cannot guarantee full compatibility between this and any other implementation.

## Proposed Solution

We should constrain the allowed padding algorithms to a whitelisted set. This does hold the risk of breaking existing usage if anyone is using an unusual padding algorithm not on this list.

### Allow

* `RSA/ECB/PKCS1Padding` - Required in the [JCE Implementation Requirements][2]
* `RSA/ECB/OAEPWithSHA-1AndMGF1Padding` - Required in the [JCE Implementation Requirements][2]
* `RSA/ECB/OAEPWithSHA-256AndMGF1Padding` - Required in the [JCE Implementation Requirements][2]
* `RSA/ECB/OAEPWithSHA-384AndMGF1Padding` - Secure and reasonably common.
* `RSA/ECB/OAEPWithSHA-512AndMGF1Padding` - Secure and already used by at least one AWS service.

### Consider for future

* `RSA/ECB/OAEPWithSHA-224AndMGF1Padding` - Better than `SHA1`, but based on consultation with algorithms team we should leave it off unless we receive requests for it.

## Alternate Solution

Alternately, we could simply document the above as officially supported padding algorithms and raise a warning if an unsupported padding algorithm is used. I like this less from a "keep the foot-guns locked away" perspective, but it is the safer option considering how long the existing implementation has been in the wild.

[1]: https://github.com/awslabs/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/jce/JceMasterKey.java#L98
[2]: https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#impl

SalusaSecondus:
This is to fix #56 

I have manually verified that all existing tests pass as well as manual verification that a logged warning appears if and only if an unsupported algorithm is used with `JceMasterKey`.

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/jce/JceMasterKey.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/jce/JceMasterKey.java
@@ -29,6 +29,8 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
 
 import javax.crypto.Cipher;
 import javax.crypto.SecretKey;
@@ -49,6 +51,7 @@ import com.amazonaws.encryptionsdk.internal.EncryptionContextSerializer;
  * {@link #getInstance(PublicKey, PrivateKey, String, String, String)}.
  */
 public abstract class JceMasterKey extends MasterKey<JceMasterKey> {
+    private static final Logger LOGGER = Logger.getLogger(JceMasterKey.class.getName());
     private static final byte[] EMPTY_ARRAY = new byte[0];
 
     private final SecureRandom rnd = new SecureRandom();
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazonaws/encryptionsdk/jce/JceMasterKey.java
```diff
@@ -234,12 +237,17 @@ public WrappingData(final Cipher cipher, final byte[] extraInfo) {
     }
 
     private static class Rsa extends JceMasterKey {
+        private static final Pattern SUPPORTED_TRANSFORMATIONS =
```
lizroth: It might be nice to have some kind of test coverage for this in general in the future to make sure we warn / don't warn as intended

PR #62, opened 2018-07-31
Make JceMasterKey Case Insensitive

johnwalker:
Algorithm names in the JCA spec are not case-sensitive. This makes
JceMasterKey algorithm names case insensitive.

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/jce/JceMasterKey.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/jce/JceMasterKey.java
@@ -72,8 +72,8 @@ public abstract class JceMasterKey extends MasterKey<JceMasterKey> {
      */
     public static JceMasterKey getInstance(final SecretKey key, final String provider, final String keyId,
             final String wrappingAlgorithm) {
-        switch (wrappingAlgorithm) {
-            case "AES/GCM/NoPadding":
+        switch (wrappingAlgorithm.toUpperCase()) {
+            case "AES/GCM/NOPADDING":
                 return new AesGcm(key, provider, keyId);
             default:
                 throw new IllegalArgumentException("Right now only AES/GCM/NoPadding is supported");
--- a/src/test/java/com/amazonaws/encryptionsdk/jce/JceMasterKeyTest.java
+++ b/src/test/java/com/amazonaws/encryptionsdk/jce/JceMasterKeyTest.java
@@ -49,22 +49,22 @@ public class JceMasterKeyTest {
      * Relies on passing an invalid algorithm name to result in an Exception.
      */
     @Test
-    public void testGetInstanceCase1() {
+    public void testGetInstanceAllLowercase() {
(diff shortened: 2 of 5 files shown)


src/test/java/com/amazonaws/encryptionsdk/jce/JceMasterKeyTest.java
```diff
@@ -0,0 +1,30 @@
...
+                "blatently/unsupported/algorithm");
+    }
+
+    /**
+     * Calls JceMasterKey.getInstance with differently cased wrappingAlgorithm names.
+     * Passes if no Exception is thrown.
+     * Relies on passing an invalid algorithm name to result in an Exception.
+     */
+    @Test
+    public void testGetInstanceCaseInsensitive() {
```
mattsb42-aws: These should be separate tests.

src/main/java/com/amazonaws/encryptionsdk/jce/JceMasterKey.java
```diff
@@ -72,8 +72,8 @@
      */
     public static JceMasterKey getInstance(final SecretKey key, final String provider, final String keyId,
             final String wrappingAlgorithm) {
-        switch (wrappingAlgorithm) {
-            case "AES/GCM/NoPadding":
+        switch (wrappingAlgorithm.toUpperCase()) {
+            case "AES/GCM/NOPADDING":
```
mattsb42-aws: GitHub won't let me comment on lines that weren't changed in the PR, but we also need to apply this type of check on line 98 where we check the `RSA/ECB/` prefix.

johnwalker: Added test case for the RSA/ECB prefix. This check is already applied

src/test/java/com/amazonaws/encryptionsdk/jce/JceMasterKeyTest.java
johnwalker: Separated the test into two cases

src/test/java/com/amazonaws/encryptionsdk/jce/JceMasterKeyTest.java
```diff
@@ -0,0 +1,70 @@
...
+        jceGetInstanceAsymmetric("rsa/ec/unsupportedAlgorithm");
+    }
+
+    /**
+     * Calls JceMasterKey.getInstance with differently cased wrappingAlgorithm names.
+     * Passes if no Exception is thrown.
+     * Relies on passing an invalid algorithm name to result in an Exception.
+     */
+    @Test
+    public void testGetInstanceCase1() {
```
mattsb42-aws: NP: Bytes are cheap. Human cognition trying to determine what a test is doing from a vague name is not.

johnwalker: Added more descriptive names.

PR #66, opened 2018-08-02
Remove the KMS client cache

bdonlan:


pom.xml
```diff
@@ -4,7 +4,7 @@
 
     <groupId>com.amazonaws</groupId>
     <artifactId>aws-encryption-sdk-java</artifactId>
-    <version>1.3.3</version>
+    <version>1.3.5-SNAPSHOT</version>
```
lizroth: I think it would be clearer to call this something other than `SNAPSHOT`, but we can figure that out later.

PR #71, opened 2018-08-28
Keyblob parse exception

slyubomirsky:
*Issue #, if available:*
No number

*Description of changes:*
Done at the request of Rustan Leino: Currently, if the KeyBlob deserializer reads a negative length for the key provider ID, key provider info, or the key itself, this does not directly result in an exception. Instead, when the deserializer later tries to read the provider ID, provider info, or key, `Arrays.copyOfRange()` throws an `IllegalArgumentException`. From a client's perspective, this seems like a misleading exception, since the client passed all the correct arguments but the message's contents are what led to the exception. This change instead throws a `ParseException` if a negative length is read, corresponding to the line "the bytes does not properly represent the desired object" in the Javadoc for `ParseException.`

Unit tests are included that test for this case. The unit tests throw an IllegalArgumentException in the present version of the code; with the changes, they terminate normally but without deserializing the field that had a negative length. (This happens because `deserialize()` catches the `ParseException`. That may not be desired behavior in that case, since a client might retry `deserialize()`, not knowing that the message contents are what led to the failure to deserialize and thus never succeeding. Perhaps `deserialize()` should distinguish between bad message bodies versus not having enough bytes with different exceptions?)

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/model/KeyBlob.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/model/KeyBlob.java
@@ -98,7 +98,11 @@ public final class KeyBlob implements EncryptedDataKey {
      *             length.
      */
     private int parseKeyProviderIdLen(final byte[] b, final int off) throws ParseException {
-        keyProviderIdLen_ = PrimitivesParser.parseShort(b, off);
+        short len = PrimitivesParser.parseShort(b, off);
+        if (len < 0) {
+            throw new ParseException("Parsed negative length for key provider id");
+        }
+        keyProviderIdLen_ = len;
         return Short.SIZE / Byte.SIZE;
     }
 
@@ -152,7 +156,11 @@ public final class KeyBlob implements EncryptedDataKey {
      *             length.
      */
     private int parseKeyProviderInfoLen(final byte[] b, final int off) throws ParseException {
-        keyProviderInfoLen_ = PrimitivesParser.parseShort(b, off);
+        short len = PrimitivesParser.parseShort(b, off);
+        if (len < 0) {
+            throw new ParseException("Parsed negative length for key provider info");
+        }
+        keyProviderInfoLen_ = len;
         return Short.SIZE / Byte.SIZE;
     }
 
@@ -205,7 +213,11 @@ public final class KeyBlob implements EncryptedDataKey {
      *             if there are not sufficient bytes to parse the key length.
      */
     private int parseKeyLen(final byte[] b, final int off) throws ParseException {
-        encryptedKeyLen_ = PrimitivesParser.parseShort(b, off);
+        short len = PrimitivesParser.parseShort(b, off);
+        if (len < 0) {
+            throw new ParseException("Parsed negative length for key");
+        }
+        encryptedKeyLen_ = len;
         return Short.SIZE / Byte.SIZE;
     }
 
--- a/src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
+++ b/src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
@@ -15,7 +15,9 @@ package com.amazonaws.encryptionsdk.model;
 
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 
+import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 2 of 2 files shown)


src/main/java/com/amazonaws/encryptionsdk/model/KeyBlob.java
```diff
@@ -98,7 +98,11 @@ public KeyBlob(final EncryptedDataKey edk) {
      *             length.
      */
     private int parseKeyProviderIdLen(final byte[] b, final int off) throws ParseException {
-        keyProviderIdLen_ = PrimitivesParser.parseShort(b, off);
+        short len = PrimitivesParser.parseShort(b, off);
```
mattsb42-aws: These methods are all doing the same thing, just with different error messages. This should be centralized into a single helper method that all three use.

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
```diff
@@ -170,4 +172,51 @@ public void checkKeyLen() {
 
         assertEquals(mockDataKey_.getEncryptedDataKey().length, reconstructedKeyBlob.getEncryptedDataKeyLen());
     }
+
```
mattsb42-aws: In addition to the direct method calls, we should have a static test vector for each of these that contains bytes that will trigger the correct error when parsed.

src/main/java/com/amazonaws/encryptionsdk/model/KeyBlob.java
slyubomirsky: Will do

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
slyubomirsky: How should these vectors be provided? As a byte array literal?

mattsb42-aws: We're working on building out more standardized approaches for these, but for the time being, yes, just some byte array literals will do nicely.

slyubomirsky: I added the vectors as functions, but wasn't sure if a field would be preferable. I would be glad to present them in whatever manner would be most useful.

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
```diff
@@ -170,4 +172,76 @@ public void checkKeyLen() {
...
+        final byte[] keyBlobBytes = createKeyBlobBytes();
+
+        // manually set the keyProviderIdLen to negative
+        final byte[] negativeKeyProviderIdLen = ByteBuffer.allocate(Short.BYTES)
+	    .putShort((short) -1).array();
+        System.arraycopy(negativeKeyProviderIdLen, 0, keyBlobBytes, 0, Short.BYTES);
+
+	// a negative field length throws a parse exception, so deserialization is incomplete
+	assertIncomplete(keyBlobBytes);
+	assertIncomplete(negativeKeyProviderIdLenTestVector());
```
mattsb42-aws: Sorry, I should have clarified. What I'm looking for is a separate test that calls the full deserialize path on a byte array crafted to contain the bad values.

That being said, the vectors you provided triggered paths in my brain that I had skipped over, that change what this issue is.

Per the message format spec[1], every one of these values is an *unsigned* integer. What would be better to be testing is to test the methods in `PrimitivesParser` to verify that none of them can result in negative values.

[1] https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/message-format.html

slyubomirsky: I see; I will investigate `PrimitivesParser` then

slyubomirsky:
Static test vectors pose more of an issue with the changes as they stand now, as they would have to be quite large to trigger the case here. Is there any particularly convenient way to provide one or another recommendation?

slyubomirsky:
(The CI appears to have failed for a permissions-related reason, not because the tests failed.)

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
```diff
@@ -170,4 +171,52 @@ public void checkKeyLen() {
...
+        assertArrayEquals(b1.getProviderId().getBytes(StandardCharsets.UTF_8),
+                          b2.getProviderId().getBytes(StandardCharsets.UTF_8));
+        assertArrayEquals(b1.getProviderInformation(), b2.getProviderInformation());
+        assertArrayEquals(b1.getEncryptedDataKey(), b2.getEncryptedDataKey());
+    }
+    
+    @Test
+    public void checkKeyProviderIdLenUnsigned() {
+        // provider id length is too large for a signed short but fits in unsigned
+        final KeyBlob blob = generateRandomKeyBlob(Short.MAX_VALUE + 1, Short.MAX_VALUE, Short.MAX_VALUE);
```
johnwalker: Why use Short.MAX_VALUE instead of the UNSIGNED_MAX_VALUE?

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
```diff
@@ -170,4 +171,52 @@ public void checkKeyLen() {
 
         assertEquals(mockDataKey_.getEncryptedDataKey().length, reconstructedKeyBlob.getEncryptedDataKeyLen());
     }
+
+    private KeyBlob generateRandomKeyBlob(int idLen, int infoLen, int keyLen) {
+        final byte[] idBytes = RandomBytesGenerator.generate(idLen);
+        // negative bytes translate into U+FFFD, so no thanks
```
SalusaSecondus: If this array is intended to be parsed as UTF-8, then you shouldn't just generate random bytes (even with some amount of checks). Why not just fill it with `'A'`?

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
```diff
@@ -170,4 +171,52 @@ public void checkKeyLen() {
...
+    public void checkKeyProviderInfoLenUnsigned() {
+        // provider info length is too large for a signed short but fits in unsigned
+        final KeyBlob blob = generateRandomKeyBlob(Short.MAX_VALUE, Short.MAX_VALUE + 2, Short.MAX_VALUE);
+        final byte[] arr = blob.toByteArray();
+
+        assertKeyBlobsEqual(deserialize(arr), blob);
+    }
+
+    @Test
+    public void checkKeyLenUnsigned() {
```
SalusaSecondus: Can you add a test for when it is too long to fit even in an unsigned short?

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
slyubomirsky: Good point

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
slyubomirsky: Good idea; I will

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
slyubomirsky: It was an arbitrary choice (just needed it to be too large to be a positive signed short but a valid unsigned short), I can change it to make the intent of the test clearer

src/test/java/com/amazonaws/encryptionsdk/model/KeyBlobTest.java
slyubomirsky: It occurs to me now that such tests already exist, namely `overlyLargeKeyProviderIdLen()`, `overlyLargeKeyProviderInfoLen()`, and `overlyLargeKey()`. They check that instantiating such a KeyBlob results in an exception. (I had changed them earlier to make sure they were treating `Constants.UNSIGNED_SHORT_MAX_VAL` as the max length.)

slyubomirsky:
Note: Technically a char could be used to store these unsigned 16-bit values, since the Java language standard guarantees that chars must be unsigned. This could be useful to have a compiler-provided guarantee that these lengths are non-negative, though admittedly a little strange.

PR #81, opened 2018-10-25
Run the OpenJML proof in Travis

seanmcl:
I tested this in a fork.

*Issue #, if available:*

*Description of changes:*

Currently there are OpenJML specifications, but they are not checked during the build.  This change changes the build to have Travis check the specs in Docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,10 @@ jdk:
 script: mvn install -Dgpg.skip=true
 sudo: false
 dist: trusty
+
+# Run the OpenJML proof only once, on openjdk8
+matrix:
+  include:
+    - stage: proof
+      jdk: openjdk8
+      script: jml/run-proof.sh
--- a/jml/run-proof.sh
+++ b/jml/run-proof.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+set -e -u -o pipefail -x
+trap "kill 0" SIGINT SIGTERM
+
+mkdir -p /var/tmp/openjml{,-build}
+
+# Install OpenJML
+pushd /var/tmp/openjml-build
+git clone https://github.com/OpenJML/OpenJML.git
+cd OpenJML && git checkout e4dc6371e2cf45d1a08b5c91990e2a875f33dd27
+cp OpenJML/releaseBuilds-CI/openjml.jar /var/tmp/openjml/openjml.jar
+popd
+
+# Install Z3
+pushd /var/tmp/openjml-build
+git clone https://github.com/OpenJML/Solvers.git
+cd Solvers && git checkout 023d9682b6fa2e5d70e7587e2e293bdeae4e0885
+cp -r Solvers-linux /var/tmp/openjml
+popd
+
+# Run the OpenJML proof.  Currently we only check the annotations on the files
+# in FILES.  We will continue adding files to this list as we add annotations.
+cd src/main/java
+
+declare -a FILES=(
+  com/amazonaws/encryptionsdk/model/KeyBlob.java
+)
+
+for file in ${FILES[@]}; do
+  java -Xmx4g -jar /var/tmp/openjml/openjml.jar -- -Werror -progress -timeout 600 -esc -spec-math=bigint -code-math=safe -minQuant -no-staticInitWarning -prover z3_4_3 -sourcepath . ${file}
+done


.travis.yml
```diff
@@ -4,6 +4,13 @@ jdk:
   - oraclejdk8
 #  - openjdk9 # Not yet supported by travis
   - oraclejdk9
-script: mvn install -Dgpg.skip=true
+services:
+  - docker
+
+script:
+  - mvn install -Dgpg.skip=true
+  - pushd . && cd jml && docker build -t openjml . && popd
```
bdonlan: Please run in the travis environment directly.

Also - does this need to run three times, on different JDKs? Ideally we'd prefer to just run this as a separate test step, instead of running it on each of the test matrix environments.

jml/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM centos:centos7
+
+RUN yum -y update
+RUN yum -y install \
+  awscli \
```
bdonlan: Are you using awscli/libgomp?

jml/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN cd OpenJML && git checkout e4dc6371e2cf45d1a08b5c91990e2a875f33dd27
+RUN cp OpenJML/OpenJML/releaseBuilds-CI/openjml.jar /usr/local/openjml/openjml.jar
+
+# Solvers
+RUN git clone https://github.com/OpenJML/Solvers.git
+RUN cd Solvers && git checkout 023d9682b6fa2e5d70e7587e2e293bdeae4e0885
+RUN cp -r Solvers/Solvers-linux /usr/local/openjml
+
+WORKDIR /usr/local/openjml
+ENTRYPOINT ["java", "-Xmx12g", "-jar", "openjml.jar", "--", "-Werror", "-progress", "-timeout", "600", "-esc", "-spec-math=bigint", "-code-math=safe", "-minQuant", "-no-staticInitWarning", "-prover", "z3_4_3"]
```
bdonlan: Does travis provide 12GB of RAM in its containers?

.travis.yml
mattsb42-aws: > Ideally we'd prefer to just run this as a separate test step, instead of running it on each of the test matrix environments.

++ Both to only run this as often as is actually needed and to separate out "tests are failing" from "spec is failing/divergent", this should be run as a separate Travis environment.

seanmcl: Yes, thanks for the feedback.  I'm not a Travis power-user, and am working on running the proof on a single JDK.  I'll try to figure out how to do it in a separate environment as well.

mattsb42-aws: Take a look at `matrix.include`. That lets you define exact combinations of configuration values for each separate environment.

https://docs.travis-ci.com/user/build-matrix/

example:

https://github.com/aws/aws-encryption-sdk-python/blob/master/.travis.yml

jml/Dockerfile
seanmcl: This is now obsolate, as we've removed docker.  (but yeah, Z3 uses libgomp)

jml/Dockerfile
seanmcl: Nice catch.  `trusty` gives 4g.  https://docs.travis-ci.com/user/reference/overview/

.travis.yml
seanmcl: Thanks.  Done.

seanmcl:
I believe I've addressed all comments.  The resulting build can be seen here:

https://travis-ci.org/seanmcl/aws-encryption-sdk-java

jml/run-proof.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+set -e -u -o pipefail -x
+trap "kill 0" SIGINT SIGTERM
+
+mkdir -p /var/tmp/openjml{,-build}
+
+# Install OpenJML
+pushd /var/tmp/openjml-build
+git clone https://github.com/OpenJML/OpenJML.git
+cd OpenJML && git checkout e4dc6371e2cf45d1a08b5c91990e2a875f33dd27
```
mattsb42-aws: This seems really brittle. What is this commit supposed to be?

seanmcl: I believe this makes the build less brittle.  When OpenJML changes, we don't want the proof to fail because it can no longer check this proof.  I can check out the tip of the branch which is supporting these builds, but our build can then break without any changes on our side.  (This is similar to builds that blindly pull from apt/yum repos.)  I'm happy to change it to pull a branch if you like, with that caveat.

bdonlan: Can we use a tag instead? This would let us do a shallow clone to reduce load on github's servers, and also make it more obvious which version we're referencing.

seanmcl:
I learned on Tuesday that we're moving away from OpenJML, so I'm dropping this PR.  I apologize for the noise.

PR #83, opened 2018-11-28
Check if the given ARN is valid before to parse the region.

tanguc:
When using the KMSMasterKeyBuilder or the default constructor, in the case where we don't provide the region 
 and only the key id (ARN), to infer the region we'll use the key id however if the key id is null there will be a NullPointerException which will be raised, and by default null(s) are not handled correctly on upstream operations.


johnwalker:
Hi! Thanks for the contribution, it looks good. Would you be willing to add a test case that demonstrates the bugfix?

acioc:
Closing stale PR, please reopen as needed

PR #97, opened 2019-02-05
add explicit public API definition

mattsb42-aws:
*Description of changes:*
Adding an explicit definition of our public API to avoid confusion.


README.md
```diff
@@ -138,6 +138,13 @@ public class StringExample {
 
 You can find more examples in the [examples directory][examples].
 
+## Public API
+
+The public API that we will maintain within the constraints of our [versioning policy](./VERSIONING.rst)
+consists of all public classes in the `com.amazonaws.encryptionsdk` package unless otherwise and specifically documented.
+
```
juneb: The public API that we maintain includes all public classes in the in the `com.amazonaws.encryptionsdk` package unless otherwise documented. Our [versioning policy](./VERSIONING.rst) applies to this public API.

johnwalker:
It might be helpful to include a short statement in the versioning policy applies to non-internal classes.

README.md
johnwalker: Edit: "Our [versioning policy](./VERSIONING.rst) applies to all public classes in the `com.amazonaws.encryptionsdk`
package unless otherwise documented."

johnwalker: Also, not to nitpick: it also applies to public constants, methods, etc.

johnwalker: ... and protected methods, constants, etc.

johnwalker: Second try: Our versioning policy applies to all public and protected classes/methods/fields in the com.amazonaws.encryptionsdk
package unless otherwise documented.

PR #109, opened 2019-05-30
Switch Base64 codec, upgrade dependencies, and bump version to 1.5.0

praus:
### Description of changes
* This change switches Base64 codec from java.util.Base64 to Base64 from Apache Commons Codec. This is to enable use of this library on systems that do not contain java.util.Base64 but do support other Java 8 features.
* Upgrade dependencies to their most recent versions.
* Bump version to 1.5.0

### Testing
* `mvn verify` passes on my laptop and runs AllTestsSuite with its 5726 tests.
* Travis CI tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.4.2 -- 2019-05-30
+## 1.5.0 -- 2019-05-30
 
 ### Minor Changes
 * Add dependency on Apache Commons Codec 1.12.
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog
 
+## 1.4.2 -- 2019-05-30
```
lizroth: Because we are adding a new dependency and doing several upgrades, I think this should probably be a minor version bump, not a patch.

https://github.com/aws/aws-encryption-sdk-java/blob/master/VERSIONING.rst

praus: Excellent point, thank you. The new revision fixes it.

Issue #114, opened 2019-06-26
ParsedCiphertext should throw an error if it is not complete


mattsb42-aws:
`ParsedCiphertext` will currently accept a partial ciphertext without error.

Before returning from the constructor, we need to check to make sure that `isComplete_` is true.

This is most problematic if the ciphertext is only one byte becase of #113.

PR #119, opened 2019-07-24
ParsedCiphertext throws an error if it is not complete

sprovine:
#114 

Added a check as well as a test to make sure `isComplete_` is true before returning from the `ParsedCiphertext` constructor.

# Check any applicable:


sprovine:
@lavaleri @josecorella

src/main/java/com/amazonaws/encryptionsdk/ParsedCiphertext.java
```diff
@@ -33,9 +35,12 @@
      * {@code ciphertext} and that any changes made to the backing array will be reflected here as
      * well.
      */
-    public ParsedCiphertext(final byte[] ciphertext) {
+    public ParsedCiphertext(final byte[] ciphertext) throws ParseException {
```
lavaleri: Because the constructor for ParsedCiphertext can now throw an error, we probably want to check everywhere ParsedCiphertext is being constructed and verify that the error will be handled appropriately, and make sure we have a test to ensure that it continues to handle in the same way. For example, we currently test the happy case for using encrypt/decrypt with ParsedCiphertext: https://github.com/aws/aws-encryption-sdk-java/blob/master/src/test/java/com/amazonaws/encryptionsdk/AwsCryptoTest.java#L199

This PR should include a new test in AwsCryptoTest to test the correct behavior of encrypt/decrypt in the case that ParsedCiphertext throws this error during construction.

src/test/java/com/amazonaws/encryptionsdk/model/CiphertextHeadersTest.java
```diff
@@ -314,6 +314,24 @@ private void readUptoNonceLen(final ByteBuffer headerBuff) {
         headerBuff.get();
     }
 
+    @Test(expected = BadCiphertextException.class)
+    public void incompleteCiphertext() {
```
lavaleri: This test is currently testing already existing behavior in the CiphertextHeader class (note that if you remove the change you made to ParsedCiphertext, this test will still pass because this test isn't handling any ParsedCiphertext). To test the change you made to ParsedCiphertext, it looks like you'll need to create a new file specifically to test ParsedCiphertext. This will also give you the opportunity to test out the happy case for ParsedCiphertext construction, which is a test that is currently missing.

src/test/java/com/amazonaws/encryptionsdk/model/CiphertextHeadersTest.java
```diff
@@ -314,6 +314,24 @@ private void readUptoNonceLen(final ByteBuffer headerBuff) {
...
+        final byte[] headerBytes = ciphertextHeaders.toByteArray();
+        final ByteBuffer headerBuff = ByteBuffer.wrap(headerBytes);
+
+        readUptoVersion(headerBuff);
+
+        // set ciphertext to incomplete.
+        headerBuff.put((byte) 0);
+
+        final CiphertextHeaders reconstructedHeaders = new CiphertextHeaders();
+        reconstructedHeaders.deserialize(headerBuff.array(), 0);
```
lavaleri: I *think* what is happening here is that this throws a BadCiphertextException because you mess up the buffer on L329. This results in BadCiphertextException not because the CiphertextHeaders is not complete, but because in the process of deserializing the bytes one of the values doesn't make sense (my guess would be here: https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/model/CiphertextHeaders.java#L205)

This means that this test is likely a duplicate of the test here: https://github.com/aws/aws-encryption-sdk-java/blob/master/src/test/java/com/amazonaws/encryptionsdk/model/CiphertextHeadersTest.java

src/main/java/com/amazonaws/encryptionsdk/ParsedCiphertext.java
```diff
@@ -1,11 +1,11 @@
 /*
  * Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
SalusaSecondus: Please update copyright.

src/main/java/com/amazonaws/encryptionsdk/ParsedCiphertext.java
SalusaSecondus: While this is unclear, I'd probably through a `BadCiphertextException` rather than a `ParseException`. This is because the latter appears to be used more for low-level primitive parsing rather than trying to understand an entire ciphertext.

Regardless, these are both descendants of `RuntimeException` and thus need-not (and likely should-not) be declared in a `throws` statement on the method.

Either way, more testing is in order.

src/main/java/com/amazonaws/encryptionsdk/ParsedCiphertext.java
```diff
@@ -33,9 +35,12 @@
      * {@code ciphertext} and that any changes made to the backing array will be reflected here as
      * well.
      */
-    public ParsedCiphertext(final byte[] ciphertext) {
+    public ParsedCiphertext(final byte[] ciphertext) throws ParseException {
         ciphertext_ = Utils.assertNonNull(ciphertext, "ciphertext");
         offset_ = deserialize(ciphertext_, 0);
+        if (!this.isComplete()) {
+          throw new BadCiphertextException("Incomplete ciphertext.");
```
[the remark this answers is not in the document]
SalusaSecondus: I respectfully disagree. While the guidance is unclear, I would stay with a `BadCiphertextException` for the reasons noted elsewhere in this PR.

src/test/java/com/amazonaws/encryptionsdk/ParsedCiphertextTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.amazonaws.encryptionsdk.exception.BadCiphertextException;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.mockito.Mockito.spy;
+
+public class ParsedCiphertextTest extends CiphertextHeaders {
+    final byte[] ciphertext_ = {0};
```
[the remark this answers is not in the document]
SalusaSecondus: Agreed. Also, please improve the name of this variable and ensure that you handle incomplete ciphertexts which are longer than a single zero byte (such as only a single byte short).

lavaleri:
Amendment to the above comment. For test (1) assert > 0 on the offset, not assertNotNull. For test (2) I was incorrect in thinking that the offset is the length of the ciphertext. It is the length of the ciphertext headers of the ciphertext you pass in. You do not need to include (2) in your next PR. For test (3), since again it is only parsing the headers of the ciphertext, you want to try to pass to it a ciphertext that *almost* has correct headers.

src/test/java/com/amazonaws/encryptionsdk/ParsedCiphertextTest.java
```diff
@@ -0,0 +1,94 @@
...
+        final Map<String, String> encryptionContext = new HashMap<String, String>(1);
+        encryptionContext.put("ENC1", "ParsedCiphertext test with %d" + byteSize);
+
+        encryptionClient_.setEncryptionFrameSize(frameSize);
+
+        final byte[] cipherText = encryptionClient_.encryptData(
+                masterKeyProvider,
+                plaintextBytes,
+                encryptionContext).getResult();
+        byte[] incompleteCipherText = Arrays.copyOfRange(cipherText, 1, cipherText.length);
```
SalusaSecondus: This isn't testing what you think because it is dropping the first byte of the header while maintaining the result of the ciphertext. This will result in a complete but invalid) header. That isn't what you are trying to test.

Try this:
```
ParsedCiphertext parsed = new ParsedCiphertext(cipherText);
byte[] incompleteCiphertext = Arrays.copyOf(parsed.getCiphertext(), parsed.getOffset() - 1);
```

src/test/java/com/amazonaws/encryptionsdk/ParsedCiphertextTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test(expected = BadCiphertextException.class)
+    public void incompleteZeroByteCiphertext() {
+        final byte[] cipherText = {};
+        ParsedCiphertext pCt = new ParsedCiphertext(cipherText);
+    }
+
+    @Test(expected = BadCiphertextException.class)
+    public void incompleteSingleByteCiphertext() {
+        final byte[] cipherText = {0};
```
lavaleri: With the fix we made to checking the version, we need to make sure this is a valid version (i.e. `VersionInfo.CURRENT_CIPHERTEXT_VERSION`).

PR #120, opened 2019-07-30
Flag deprecated encryptString/decryptString APIs as deprecated.

lizroth:
*Issue #, if available:* n/a

*Description of changes:*

* Add `@Deprecated` to deprecated APIs
* Update Javadoc about deprecation
* Opportunistic improvement of incomplete Javadoc on decryptData()

**Background:** the `encryptString`/`decryptString` APIs were an early approach to try to provide helpful encoding for string ciphertexts. However, the AWS Encryption SDKs in general attempt to be agnostic as to the data format that users want to use for their ciphertexts, so over the longer term these APIs have turned out to be more confusing than beneficial for customers. As such, most Encryption SDKs do not support these APIs by design.

Only the Java ESDK has these APIs, because Java was the original implementation, and at the time it was a more significant convenience improvement to offer these APIs to Java developers.

This PR deprecates those APIs to make it clear to our users which APIs are interoperable and to encourage our users to prefer the ones that are.

# Check any applicable:


src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -281,7 +281,11 @@ public long estimateCiphertextSize(
     /**
      * Calls {@link #encryptData(MasterKeyProvider, byte[], Map)} on the UTF-8 encoded bytes of
      * {@code plaintext} and base64 encodes the result.
+     * @deprecated Use the {@link #encryptData(MasterKeyProvider, byte[], Map)} and
+     * {@link #decryptData(MasterKeyProvider, byte[])} APIs instead. {@code encryptString} is not out-of-the-box
+     * compatible with the other Encryption SDKs and intentionally not supported by the other implementations.
      */
```
juneb: {@code encryptString} is incompatible with the other AWS Encryption SDK language implementations. By design, this feature is not supported by the other language implementations.

lizroth: Thanks, fixing.

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -281,7 +281,12 @@ public long estimateCiphertextSize(
     /**
      * Calls {@link #encryptData(MasterKeyProvider, byte[], Map)} on the UTF-8 encoded bytes of
      * {@code plaintext} and base64 encodes the result.
+     * @deprecated Use the {@link #encryptData(MasterKeyProvider, byte[], Map)} and
+     * {@link #decryptData(MasterKeyProvider, byte[])} APIs instead. {@code encryptString} is incompatible with the
```
mattsb42-aws: I would not say that they are incompatible without clarification. Maybe "because the the output must be base64-decoded to be compatible with other AWS Encryption SDK implementations"?

juneb:
@lizroth We should also add the deprecation message to the example that demonstrates how to use the encryptString and decryptString methods, or remove them.

https://github.com/aws/aws-encryption-sdk-java/blob/master/src/examples/java/com/amazonaws/crypto/examples/StringExample.java

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
lizroth: Works for me, I am but a deprecation-shaped vessel for the wordsmithing that you guys think make sense.

juneb: Because it returns Base64-encoded ciphertext, this API is not fully compatible with other AWS Encryption SDK language implementations.

mattsb42-aws: > Because these APIs return Base64-encoded output, they are not compatible with other AWS Encryption SDK language implementations.

IMO this still conveys the wrong message.
What I would like to make sure we're communicating is that they *are* compatible, *but* only once you peal away an extra layer of encoding that this method adds.

lizroth: After some noodling, I've tried again to be clear and concise in explaining that this is deprecated, and why. When you have a second, please let me know what you think of the new wording.

lizroth:
> @lizroth We should also add the deprecation message to the example that demonstrates how to use the encryptString and decryptString methods, or remove them.
> 
> https://github.com/aws/aws-encryption-sdk-java/blob/master/src/examples/java/com/amazonaws/crypto/examples/StringExample.java

Agreed. Removed the example.

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -281,7 +281,12 @@ public long estimateCiphertextSize(
     /**
      * Calls {@link #encryptData(MasterKeyProvider, byte[], Map)} on the UTF-8 encoded bytes of
      * {@code plaintext} and base64 encodes the result.
+     * @deprecated Use the {@link #encryptData(MasterKeyProvider, byte[], Map)} and
+     * {@link #decryptData(MasterKeyProvider, byte[])} APIs instead. {@code encryptString} and {@code decryptString}
```
seebees: I would break on the `.`

juneb: @deprecated Use the {@link #encryptData(MasterKeyProvider, byte[], Map)} and
{@link #decryptData(MasterKeyProvider, byte[])} APIs instead. {@code encryptString} and {@code decryptString} 
work as expected if you use them together. However, to work with other language implementations of the AWS Encryption  SDK, you need to base64 decode the output of {@code encryptString} and base64-encode the input to 
{@code decryptString}. These deprecated APIs will be removed in the future.

Issue #41, opened 2017-11-28
Remove hard cryptographic dependency on BouncyCastle


SalusaSecondus:
Right now this code depends on BouncyCastle for several cases:

- [x]  EC Key generation (to support next point)
- [x] EC Key (de)serialization with point-compression
- [x] RSA encryption/decryption (due to default java JCE problems with OAEP not using SHA-1)
- [x] ECDSA signatures
- [x] HMAC-based Extract-and-Expand Key Derivation Function

To make this library more portable, we should remove all of these hard dependencies. This will require:
* Implementing point (de)compression
* Properly using AlgorithmParameterSpec to configure the JCE provider to do the right OAEP padding
* Replace HKDFBytesGenerator with Hkdf.java from aws-dynamodb-encryption-java
* Remove all other references to BouncyCastle.

fieldju:
What is the status of this issue?

fieldju:
@mattsb42-aws what is the status of this issue?

mattsb42-aws:
Adding here for reference: when we do the point decompression, we need to make sure that we cover this too: https://github.com/aws/aws-encryption-sdk-python/issues/113

ragona:
@fieldju We're taking a look at this one right now. I'll post again when we have a more concrete update, but I wanted to let you know that we're on it!

SalusaSecondus:
Confirmed that all hard cryptographic dependencies on BouncyCastle  have been removed. This means that while BouncyCastle is still being used, it is only for parsing and support logic.

Critically, it means that it can also be swapped out for the FIPS validated provider. The following diff still passes all tests:
```
diff --git a/pom.xml b/pom.xml
index 3a2f2d3..ce56b08 100644
--- a/pom.xml
+++ b/pom.xml
@@ -48,8 +48,8 @@

         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcprov-ext-jdk15on</artifactId>
-            <version>1.61</version>
+            <artifactId>bc-fips</artifactId>
+            <version>1.0.1</version>
         </dependency>

         <dependency>
```

Excellent work @WesleyRosenblum !

PR #131, opened 2019-10-18
Remaining fixes to allow BouncyCastle to be swapped out with other implementations.

WesleyRosenblum:
*Issue #, if available:* #41

*Description of changes:*

Remaining fixes to allow BouncyCastle to be swapped out with other implementations.

# Check any applicable:


src/main/java/com/amazonaws/encryptionsdk/internal/Utils.java
```diff
@@ -268,7 +268,7 @@ public static ByteBuffer limit(final ByteBuffer buff, final int newLimit) {
      * @return decoded data as a byte array
      */
     public static byte[] decodeBase64String(final String encoded) {
-        return Base64.decode(encoded);
+        return encoded.isEmpty() ? new byte[0] : Base64.decode(encoded);
```
SalusaSecondus: We should already have a constant defined for a zero-byte array. Return that instead. Otherwise you create needless memory pressure.

src/test/java/com/amazonaws/encryptionsdk/jce/KeyStoreProviderTest.java
```diff
@@ -46,8 +43,15 @@
...
 import com.amazonaws.encryptionsdk.exception.CannotUnwrapDataKeyException;
 import com.amazonaws.encryptionsdk.multi.MultipleProviderFactory;
+import sun.security.x509.AlgorithmId;
+import sun.security.x509.CertificateAlgorithmId;
+import sun.security.x509.CertificateSerialNumber;
+import sun.security.x509.CertificateValidity;
+import sun.security.x509.CertificateX509Key;
+import sun.security.x509.X500Name;
+import sun.security.x509.X509CertImpl;
+import sun.security.x509.X509CertInfo;
```
SalusaSecondus: Is there any way to do this without importing `sun.` packages. This may be test code but it's a pretty severe anti-pattern.

src/main/java/com/amazonaws/encryptionsdk/internal/Utils.java
WesleyRosenblum: I didn't see a constant defined in Constants.java, so I just used the one in Apache Commons Lang ArrayUtils. If you prefer I can add the constant to Constants.java

src/test/java/com/amazonaws/encryptionsdk/jce/KeyStoreProviderTest.java
WesleyRosenblum: Standard BC and FIPS BC have different APIs for certificate generation, and the two packages are not compatible with each other so we can't just add a new test dependency. I added a comment about the usage of the sun packages.

Issue #102, opened 2019-04-05
add keyrings


mattsb42-aws:
Starting with the AWS Encryption SDK for C, we moved from master keys/providers to keyrings as a configuration/framework primitive. We need to add corresponding keyrings to this client as well.

See #150 for a discussion on the evolution of the API as we transition to Keyrings
See #153 for a discussion on the ergonomics of Keyrings

All keyring development PRs should be made to the `keyring` branch.

- [x] Create keyring trace and add to encryption and decryption materials #134
- [x] Keyring interface #142 
- [x] KMS keyring  #147 
- [x] Raw RSA keyring #142 
- [x] Raw AES keyring #142 
- [x] Multi keyring #148
- [x] ~~MasterKeyProvider keyring #149~~
- [x] ~Master Key Provider Keyring Spec: [#62](https://github.com/awslabs/aws-encryption-sdk-specification/pull/62)~
- [x] Incorporate Keyrings into AwsCrypto and deprecate MasterKeyProviders #151 
- [x] New example code for Keyrings #155
- [x] Ensure new JUnit 5 tests for Keyrings are being run as part of Travis #154 
- [x] Update Readme example code #156
- [ ] Update Changelog with new deprecation notice and Keyring features #174 
- [x] Merge to master branch #173
- [ ] Release to Maven

PR #134, opened 2019-10-25
Create keyring trace and add to encryption and decryption materials.

WesleyRosenblum:
*Issue #, if available:* #102

*Description of changes:*

Creating a keyring trace and adding to encryption and decryption materials to allow for auditing actions a keyring has taken on encryption materials.

# Check any applicable:

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/keyrings/KeyringTraceFlag.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/keyrings/KeyringTraceFlag.java
@@ -22,28 +22,28 @@ public enum KeyringTraceFlag {
     /**
      * A flag to represent that a keyring has generated a plaintext data key.
      */
-    WRAPPING_KEY_GENERATED_DATA_KEY,
+    GENERATED_DATA_KEY,
 
     /**
      * A flag to represent that a keyring has created an encrypted data key.
      */
-    WRAPPING_KEY_ENCRYPTED_DATA_KEY,
+    ENCRYPTED_DATA_KEY,
 
     /**
      * A flag to represent that a keyring has obtained the
      * corresponding plaintext data key from an encrypted data key.
      */
-    WRAPPING_KEY_DECRYPTED_DATA_KEY,
+    DECRYPTED_DATA_KEY,
 
     /**
      * A flag to represent that the keyring has cryptographically
      * bound the encryption context to a newly created encrypted data key.
      */
-    WRAPPING_KEY_SIGNED_ENC_CTX,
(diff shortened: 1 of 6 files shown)


src/test/java/com/amazonaws/encryptionsdk/keyrings/KeyringTraceTest.java
```diff
@@ -0,0 +1,50 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package com.amazonaws.encryptionsdk.keyrings;
+
+import org.junit.Test;
+
+import static java.util.Collections.singleton;
+import static org.junit.Assert.assertEquals;
+
```
mattsb42-aws: Can we get some tests of the equality checks and tostrings?

src/main/java/com/amazonaws/encryptionsdk/keyrings/KeyringTraceFlag.java
```diff
@@ -0,0 +1,49 @@
...
+     * bound the encryption context to a newly created encrypted data key.
+     */
+    WRAPPING_KEY_SIGNED_ENC_CTX,
+
+    /**
+     * A flag to represent that the keyring has verified that an encrypted
+     * data key was originally created with a particular encryption context.
+     */
+    WRAPPING_KEY_VERIFIED_ENC_CTX
+}
```
mattsb42-aws: Thoughts on changing these names to spell out "encryption context"? I know these are the magic strings that we've used to date, but since we decided we're not worrying about communicating them across implementations I kind of want to go back in the Python implementation and expand this.

src/test/java/com/amazonaws/encryptionsdk/keyrings/KeyringTraceTest.java
WesleyRosenblum: Added tests for equality. For toString, I'm a little hesitant to add tests, mainly because I'm using a library to generate the toString, so it's not really guaranteed (nor particularly important) how the formatting will end up if the library is updated.

src/main/java/com/amazonaws/encryptionsdk/keyrings/KeyringTraceFlag.java
WesleyRosenblum: Sounds good to me. And actually the spec already spells out the word anyway (https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/structures.md#flags). I'll get rid of the "WRAPPING_KEY_" part too, I don't think that is adding much

PR #142, opened 2019-11-11
Defining Keyring interface, RawAesKeyring and RawRsaKeyring.

WesleyRosenblum:
*Issue #, if available:* #102

*Description of changes:*

This change defines the Keyring interface, RawAesKeyring and RawRsaKeyring.

# Check any applicable:

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java
@@ -28,7 +28,6 @@ import javax.crypto.SecretKey;
 
 import com.amazonaws.encryptionsdk.CryptoAlgorithm;
 import com.amazonaws.encryptionsdk.CryptoMaterialsManager;
-import com.amazonaws.encryptionsdk.DataKey;
 import com.amazonaws.encryptionsdk.DefaultCryptoMaterialsManager;
 import com.amazonaws.encryptionsdk.MasterKey;
 import com.amazonaws.encryptionsdk.MasterKeyProvider;
@@ -61,7 +60,8 @@ public class DecryptionHandler<K extends MasterKey<K>> implements MessageCryptoH
 
     private CryptoHandler contentCryptoHandler_;
 
-    private DataKey<K> dataKey_;
+    private SecretKey dataKey_;
+    private K masterKey_;
     private SecretKey decryptionKey_;
     private CryptoAlgorithm cryptoAlgo_;
     private Signature trailingSig_;
@@ -454,12 +454,13 @@ public class DecryptionHandler<K extends MasterKey<K>> implements MessageCryptoH
 
         DecryptionMaterials result = materialsManager_.decryptMaterials(request);
 
+        dataKey_ = result.getCleartextDataKey();
         //noinspection unchecked
-        dataKey_ = (DataKey<K>)result.getDataKey();
+        masterKey_ = (K)result.getMasterKey();
         PublicKey trailingPublicKey = result.getTrailingSignatureKey();
 
         try {
-            decryptionKey_ = cryptoAlgo_.getEncryptionKeyFromDataKey(dataKey_.getKey(), ciphertextHeaders);
+            decryptionKey_ = cryptoAlgo_.getEncryptionKeyFromDataKey(dataKey_, ciphertextHeaders);
         } catch (final InvalidKeyException ex) {
             throw new AwsCryptoException(ex);
         }
@@ -536,7 +537,7 @@ public class DecryptionHandler<K extends MasterKey<K>> implements MessageCryptoH
 
     @Override
     public List<K> getMasterKeys() {
-        return Collections.singletonList(dataKey_.getMasterKey());
+       return Collections.singletonList(masterKey_);
     }
 
     @Override
--- a/src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java
@@ -28,6 +28,7 @@ import javax.crypto.SecretKey;
 
 import com.amazonaws.encryptionsdk.CryptoAlgorithm;
 import com.amazonaws.encryptionsdk.CryptoMaterialsManager;
+import com.amazonaws.encryptionsdk.DataKey;
 import com.amazonaws.encryptionsdk.DefaultCryptoMaterialsManager;
 import com.amazonaws.encryptionsdk.MasterKey;
 import com.amazonaws.encryptionsdk.MasterKeyProvider;
@@ -60,8 +61,7 @@ public class DecryptionHandler<K extends MasterKey<K>> implements MessageCryptoH
 
     private CryptoHandler contentCryptoHandler_;
 
-    private SecretKey dataKey_;
-    private K masterKey_;
+    private DataKey<K> dataKey_;
     private SecretKey decryptionKey_;
     private CryptoAlgorithm cryptoAlgo_;
     private Signature trailingSig_;
@@ -454,13 +454,12 @@ public class DecryptionHandler<K extends MasterKey<K>> implements MessageCryptoH
 
         DecryptionMaterials result = materialsManager_.decryptMaterials(request);
 
-        dataKey_ = result.getCleartextDataKey();
         //noinspection unchecked
-        masterKey_ = (K)result.getMasterKey();
+        dataKey_ = (DataKey<K>)result.getDataKey();
         PublicKey trailingPublicKey = result.getTrailingSignatureKey();
 
         try {
-            decryptionKey_ = cryptoAlgo_.getEncryptionKeyFromDataKey(dataKey_, ciphertextHeaders);
+            decryptionKey_ = cryptoAlgo_.getEncryptionKeyFromDataKey(dataKey_.getKey(), ciphertextHeaders);
         } catch (final InvalidKeyException ex) {
             throw new AwsCryptoException(ex);
         }
@@ -537,7 +536,7 @@ public class DecryptionHandler<K extends MasterKey<K>> implements MessageCryptoH
 
     @Override
     public List<K> getMasterKeys() {
-       return Collections.singletonList(masterKey_);
+        return Collections.singletonList(dataKey_.getMasterKey());
     }
 
     @Override
--- a/src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
@@ -38,6 +38,7 @@ public final class DecryptionMaterials {
     private DecryptionMaterials(Builder b) {
         notNull(b.algorithmSuite, "algorithmSuite is required");
         notNull(b.keyringTrace, "keyringTrace is required");
+        notNull(b.encryptionContext, "encryptionContext is required");
         validatePlaintextDataKey(b.algorithmSuite, b.plaintextDataKey);
         validateVerificationKey(b.algorithmSuite, b.verificationKey);
 
@@ -55,7 +56,25 @@ public final class DecryptionMaterials {
         return algorithmSuite;
     }
 
-    public SecretKey getPlaintextDataKey() {
+    /**
+     * Returns true if a plaintext data key has been populated.
+     *
+     * @return True if plaintext data key is populated, false otherwise.
+     */
+    public boolean hasPlaintextDataKey() {
+        return this.plaintextDataKey != null;
+    }
+
+    /**
+     * A data key to be used as input for encryption.
+     *
+     * @return The plaintext data key.
+     * @throws IllegalStateException if plaintext data key has not been populated.
+     */
+    public SecretKey getPlaintextDataKey() throws IllegalStateException {
+        if (!hasPlaintextDataKey()) {
+            throw new IllegalStateException("plaintextDataKey has not been populated");
+        }
         return plaintextDataKey;
     }
 
@@ -66,7 +85,7 @@ public final class DecryptionMaterials {
      * @param keyringTraceEntry The keyring trace entry recording this action.
      */
     public void setPlaintextDataKey(SecretKey plaintextDataKey, KeyringTraceEntry keyringTraceEntry) {
-        if (this.plaintextDataKey != null) {
+        if (hasPlaintextDataKey()) {
             throw new IllegalStateException("plaintextDataKey was already populated");
         }
         notNull(plaintextDataKey, "plaintextDataKey is required");
@@ -76,7 +95,26 @@ public final class DecryptionMaterials {
         keyringTrace.add(keyringTraceEntry);
     }
 
-    public PublicKey getVerificationKey() {
+    /**
+     * Returns true if verification key has been populated.
+     *
+     * @return True if verification key is populated, false otherwise.
+     */
+    public boolean hasVerificationKey() {
+        return verificationKey != null;
+    }
+
+    /**
+     * The verification key used for signature verification.
+     *
+     * @return The verification key.
+     * @throws IllegalStateException if a verification key has not been populated.
+     */
+    public PublicKey getVerificationKey() throws IllegalStateException {
+        if (!hasVerificationKey()) {
+            throw new IllegalStateException(String.format(
+                    "Signature verification is not supported by AlgorithmSuite %s", algorithmSuite.name()));
+        }
         return verificationKey;
     }
 
@@ -125,7 +163,7 @@ public final class DecryptionMaterials {
     }
 
     /**
-     * Validates that a verification key is specified only if and only if
+     * Validates that a verification key is specified if and only if
      * the given algorithm suite supports signature verification.
      */
     private void validateVerificationKey(CryptoAlgorithm algorithmSuite, PublicKey verificationKey) throws IllegalArgumentException {
@@ -142,7 +180,7 @@ public final class DecryptionMaterials {
         private CryptoAlgorithm algorithmSuite;
         private SecretKey plaintextDataKey;
         private PublicKey verificationKey;
-        private Map<String, String> encryptionContext;
+        private Map<String, String> encryptionContext = Collections.emptyMap();
         private KeyringTrace keyringTrace = new KeyringTrace();
 
         private Builder(CryptoAlgorithm algorithmSuite) {
@@ -150,11 +188,11 @@ public final class DecryptionMaterials {
         }
 
         private Builder(DecryptionMaterials result) {
-            this.algorithmSuite = result.getAlgorithmSuite();
-            this.plaintextDataKey = result.getPlaintextDataKey();
-            this.verificationKey = result.getVerificationKey();
-            this.encryptionContext = result.getEncryptionContext();
-            this.keyringTrace = result.getKeyringTrace();
+            this.algorithmSuite = result.algorithmSuite;
+            this.plaintextDataKey = result.plaintextDataKey;
+            this.verificationKey = result.verificationKey;
+            this.encryptionContext = result.encryptionContext;
+            this.keyringTrace = result.keyringTrace;
         }
 
         public Builder setAlgorithmSuite(CryptoAlgorithm algorithmSuite) {
--- a/src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
@@ -62,11 +62,12 @@ public final class DecryptionMaterials {
 
     /**
      * Sets the plaintext data key. The plaintext data key must not already be populated.
-     * @param plaintextDataKey The plaintext data key.
+     *
+     * @param plaintextDataKey  The plaintext data key.
      * @param keyringTraceEntry The keyring trace entry recording this action.
      */
     public void setPlaintextDataKey(SecretKey plaintextDataKey, KeyringTraceEntry keyringTraceEntry) {
-        if(this.plaintextDataKey != null) {
+        if (this.plaintextDataKey != null) {
             throw new IllegalStateException("plaintextDataKey was already populated");
         }
         notNull(plaintextDataKey, "plaintextDataKey is required");
@@ -114,7 +115,7 @@ public final class DecryptionMaterials {
     }
 
     private void validatePlaintextDataKey(CryptoAlgorithm algorithm, SecretKey plaintextDataKey) throws IllegalArgumentException {
-        if(plaintextDataKey != null) {
+        if (plaintextDataKey != null) {
             isTrue(algorithm.getDataKeyLength() == plaintextDataKey.getEncoded().length,
                     String.format("Incorrect key length. Expected %s but got %s",
                             algorithm.getDataKeyLength(), plaintextDataKey.getEncoded().length));
--- a/src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
@@ -29,19 +29,19 @@ import static org.apache.commons.lang3.Validate.notNull;
  * Contains the cryptographic materials needed for a decryption operation with Keyrings.
  */
 public final class DecryptionMaterials {
-    private final CryptoAlgorithm algorithm;
+    private final CryptoAlgorithm algorithmSuite;
     private SecretKey plaintextDataKey;
     private final PublicKey verificationKey;
     private final Map<String, String> encryptionContext;
     private final KeyringTrace keyringTrace;
 
     private DecryptionMaterials(Builder b) {
-        notNull(b.algorithm, "algorithm is required");
+        notNull(b.algorithmSuite, "algorithmSuite is required");
         notNull(b.keyringTrace, "keyringTrace is required");
-        validatePlaintextDataKey(b.algorithm, b.plaintextDataKey);
-        validateVerificationKey(b.algorithm, b.verificationKey);
+        validatePlaintextDataKey(b.algorithmSuite, b.plaintextDataKey);
+        validateVerificationKey(b.algorithmSuite, b.verificationKey);
 
-        algorithm = b.algorithm;
+        algorithmSuite = b.algorithmSuite;
         plaintextDataKey = b.plaintextDataKey;
         verificationKey = b.verificationKey;
         encryptionContext = b.encryptionContext;
@@ -49,10 +49,10 @@ public final class DecryptionMaterials {
     }
 
     /**
-     * The algorithm to use for this decryption operation.
+     * The algorithm suite to use for this decryption operation.
      */
-    public CryptoAlgorithm getAlgorithm() {
-        return algorithm;
+    public CryptoAlgorithm getAlgorithmSuite() {
+        return algorithmSuite;
     }
 
     public SecretKey getPlaintextDataKey() {
@@ -71,7 +71,7 @@ public final class DecryptionMaterials {
         }
         notNull(plaintextDataKey, "plaintextDataKey is required");
         notNull(keyringTraceEntry, "keyringTraceEntry is required");
-        validatePlaintextDataKey(algorithm, plaintextDataKey);
+        validatePlaintextDataKey(algorithmSuite, plaintextDataKey);
         this.plaintextDataKey = plaintextDataKey;
         keyringTrace.add(keyringTraceEntry);
     }
@@ -93,7 +93,7 @@ public final class DecryptionMaterials {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         DecryptionMaterials that = (DecryptionMaterials) o;
-        return algorithm == that.algorithm &&
+        return algorithmSuite == that.algorithmSuite &&
                 Objects.equals(plaintextDataKey, that.plaintextDataKey) &&
                 Objects.equals(verificationKey, that.verificationKey) &&
                 Objects.equals(encryptionContext, that.encryptionContext) &&
@@ -102,7 +102,7 @@ public final class DecryptionMaterials {
 
     @Override
     public int hashCode() {
-        return Objects.hash(algorithm, plaintextDataKey, verificationKey, encryptionContext, keyringTrace);
+        return Objects.hash(algorithmSuite, plaintextDataKey, verificationKey, encryptionContext, keyringTrace);
     }
 
     public static Builder newBuilder(CryptoAlgorithm algorithm) {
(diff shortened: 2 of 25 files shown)


WesleyRosenblum:
Some debate on the visibility of RawAesKeyring and RawRsaKeyring is contained here: https://github.com/aws/aws-encryption-sdk-java/commit/4bbe029f34ff270d4f2b7d9c3b232d092dd6af4e#r35839832

src/main/java/com/amazonaws/encryptionsdk/keyrings/Keyring.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Constructs a {@code Keyring} which does local AES-GCM encryption
+     * decryption of data keys using the provided wrapping key.
+     *
+     * @param keyNamespace A UTF-8 encoded value that, together with the key name, identifies the wrapping key.
+     * @param keyName      A UTF-8 encoded value that, together with the key namespace, identifies the wrapping key.
+     * @param wrappingKey  The AES key input to AES-GCM to encrypt plaintext data keys.
+     * @return The {@link Keyring}
+     */
+    static Keyring rawAes(String keyNamespace, String keyName, SecretKey wrappingKey) {
```
SalusaSecondus: As noted in other discussion, please move factory methods off of this interface.

src/main/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyring.java
```diff
@@ -0,0 +1,67 @@
...
+
+        return true;
+    }
+
+    @Override
+    void traceOnEncrypt(KeyringTrace keyringTrace) {
+        keyringTrace.add(keyNamespace, keyName,
+                KeyringTraceFlag.ENCRYPTED_DATA_KEY,
+                KeyringTraceFlag.SIGNED_ENCRYPTION_CONTEXT,
+                KeyringTraceFlag.VERIFIED_ENCRYPTION_CONTEXT);
```
SalusaSecondus: Why is "VERIFIED" set on Encrypt?

src/main/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyring.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    void traceOnDecrypt(KeyringTrace keyringTrace) {
+        keyringTrace.add(keyNamespace, keyName,
+                KeyringTraceFlag.DECRYPTED_DATA_KEY,
+                KeyringTraceFlag.SIGNED_ENCRYPTION_CONTEXT,
```
SalusaSecondus: Why is "SIGNED" set on decrypt?

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyringTest.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.charset.StandardCharsets;
+
+import static com.amazonaws.encryptionsdk.keyrings.RawKeyringTest.KEYNAME;
+import static com.amazonaws.encryptionsdk.keyrings.RawKeyringTest.KEYNAMESPACE;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RawAesKeyringTest {
```
SalusaSecondus: Please add explicit test for the generate case

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawKeyringTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RawKeyringTest {
```
SalusaSecondus: I admit this is me not fully trusting mocks, but I'd like to see at least some of these tests duplicated with either (or both) of the AES/RSA keyring tests. Those could extend an abstract base test case which implements most of the tests in a generic way.

The problem is that mocks work exactly how we expect/tell them to rather than how reality might. It's important to know that the logic below works when actually implemented in addition to at the abstract level when mocked out.

src/main/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyring.java
WesleyRosenblum: I misunderstood [this issue](https://github.com/awslabs/aws-encryption-sdk-specification/issues/32), I'll fix it

src/main/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyring.java
WesleyRosenblum: Same as above

src/main/java/com/amazonaws/encryptionsdk/keyrings/Keyring.java
WesleyRosenblum: I'll move these to a StandardKeyrings class instead

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyringTest.java
WesleyRosenblum: Added

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawKeyringTest.java
WesleyRosenblum: I've added some tests to the RawAesKeyringTest and RawRsaKeyringTest that work without mocks

src/main/java/com/amazonaws/encryptionsdk/model/DecryptionMaterials.java
```diff
@@ -37,24 +78,47 @@ public Builder toBuilder() {
...
+            return this;
+        }
+
+        public SecretKey getCleartextDataKey() {
+            return cleartextDataKey;
         }
 
-        public Builder setDataKey(DataKey<?> dataKey) {
-            this.dataKey = dataKey;
+        public Builder setCleartextDataKey(SecretKey cleartextDataKey) {
```
seebees: the other implementations call this the 'unencrypted data key' and the spec uses plaintext.  I would pick one of these :)

At this point, the materials should verify that the data key conforms to algorithm specification.
The encrypt materials make sure that this can not be set more than once.
I would do that here as well.

These set methods generally also build the Keyring Trace.
This makes it harder to return materials with incorrect keyring traces.

src/main/java/com/amazonaws/encryptionsdk/model/DecryptionMaterials.java
```diff
@@ -37,24 +78,47 @@ public Builder toBuilder() {
...
+            return algorithm;
+        }
+
+        public Builder setAlgorithm(CryptoAlgorithm algorithm) {
+            this.algorithm = algorithm;
+            return this;
+        }
+
+        public SecretKey getCleartextDataKey() {
+            return cleartextDataKey;
```
seebees: Not a java expert :) . Can someone who gets this value manipulate it and change it for subsequent callers?  Or is this `final`?

src/main/java/com/amazonaws/encryptionsdk/model/EncryptionMaterials.java
```diff
@@ -75,6 +75,14 @@ public SecretKey getCleartextDataKey() {
         return cleartextDataKey;
     }
 
+    public void setCleartextDataKey(SecretKey cleartextDataKey) {
+        if(this.cleartextDataKey != null) {
```
seebees: I would validate the length, similar to the decrypt comment above.

src/main/java/com/amazonaws/encryptionsdk/model/DecryptionMaterials.java
WesleyRosenblum: SecretKeySpec, which implements the SecretKey interface, clones the encoded key byte array before returning it, so it should be ok:

```java
public byte[] getEncoded() {
    return (byte[])this.key.clone();
}
```

src/main/java/com/amazonaws/encryptionsdk/model/DecryptionMaterials.java
WesleyRosenblum: > the other implementations call this the 'unencrypted data key' and the spec uses plaintext. I would pick one of these :)

I'll go with plaintextDataKey to match the spec

>At this point, the materials should verify that the data key conforms to algorithm specification.

What do you mean by "At this point"? During the construction of DecryptionMaterials? The plaintextDataKey would be null at this point, it is set in onDecrypt in the Keyring, which uses the Algorithm in the DecryptionMaterials to create the plaintextDataKey:

```java
decryptionMaterials.setCleartextDataKey(
                            new SecretKeySpec(decryptedKey, decryptionMaterials.getAlgorithm().getDataKeyAlgo()));
                    traceOnDecrypt(decryptionMaterials.getKeyringTrace());
```

>The encrypt materials make sure that this can not be set more than once.
I would do that here as well.

This is just the builder, the actual setter for setCleartextDataKey throws an IllegalStateException if it has already been set. 

>These set methods generally also build the Keyring Trace.
This makes it harder to return materials with incorrect keyring traces.

I went back and forth on that, I wasn't happy with how it turned out when I did it that way, but let me see what I can do

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyringTest.java
```diff
@@ -0,0 +1,125 @@
...
+        EncryptionMaterials encryptionMaterials = EncryptionMaterials.newBuilder()
+                .setAlgorithm(ALGORITHM)
+                .setCleartextDataKey(DATA_KEY)
+                .setKeyringTrace(new KeyringTrace())
+                .setEncryptionContext(ENCRYPTION_CONTEXT)
+                .build();
+
+        keyring.onEncrypt(encryptionMaterials);
+
+        assertEquals(1, encryptionMaterials.getEncryptedDataKeys().size());
```
lavaleri: RawAES/RSAKeyrings have different behavior for creating edks (specifically how the AES Keyring serializes the provider Info), so I think we should explicitly test that behavior in the RawAES/RSAKeyring tests by checking the fields of the returned EDKs.

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyringTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertEquals(1, encryptionMaterials.getKeyringTrace().getEntries().size());
+        assertEquals(KEYNAME, encryptionMaterials.getKeyringTrace().getEntries().get(0).getKeyName());
+        assertEquals(KEYNAMESPACE, encryptionMaterials.getKeyringTrace().getEntries().get(0).getKeyNamespace());
+        assertEquals(2, encryptionMaterials.getKeyringTrace().getEntries().get(0).getFlags().size());
+        assertTrue(encryptionMaterials.getKeyringTrace().getEntries().get(0).getFlags().contains(KeyringTraceFlag.ENCRYPTED_DATA_KEY));
+        assertTrue(encryptionMaterials.getKeyringTrace().getEntries().get(0).getFlags().contains(KeyringTraceFlag.SIGNED_ENCRYPTION_CONTEXT));
```
lavaleri: This might be overkill, but we can also check that this is the only trace, and that it does not include the "GENERATED" flag.

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawKeyringTest.java
```diff
@@ -0,0 +1,221 @@
...
+                .build();
+
+        keyring.onDecrypt(decryptionMaterials, Collections.singletonList(ENCRYPTED_DATA_KEY));
+
+        assertEquals(DATA_KEY, decryptionMaterials.getCleartextDataKey());
+        assertEquals(0, decryptionMaterials.getKeyringTrace().getEntries().size());
+    }
+
+    @Test
+    public void testDecryptNoValidDataKey() {
```
lavaleri: The behavior is the same, but I would also add a test for when given an empty EDK list.

src/main/java/com/amazonaws/encryptionsdk/model/DecryptionMaterials.java
lavaleri: +1 on using plaintextDataKey.

> The plaintextDataKey would be null at this point, it is set in onDecrypt in the Keyring, which uses the Algorithm in the DecryptionMaterials to create the plaintextDataKey:

I think that we want to ensure that anyone who builds DecryptionMaterials, if they set a plaintextDataKey, can only successfully do so if the plaintextDataKey being set matches the algorithmSuite already set in the decryptionMaterials. i.e. let's make it harder for custom keyrings to create and return a set of materials which are not valid.

src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java
```diff
@@ -61,7 +60,8 @@
 
     private CryptoHandler contentCryptoHandler_;
 
-    private DataKey<K> dataKey_;
+    private SecretKey dataKey_;
```
lavaleri: s/dataKey_/plaintextDataKey_

src/main/java/com/amazonaws/encryptionsdk/model/DecryptionMaterials.java
WesleyRosenblum: Agreed, I'm adding validation to DecryptionMaterials and EncryptionMaterials in both the setPlaintextDataKey and in the Builder.build methods so that it will be impossible to construct an invalid DecryptionMaterials/EncryptionMaterials

src/main/java/com/amazonaws/encryptionsdk/keyrings/RawKeyring.java
```diff
@@ -0,0 +1,136 @@
...
+                            new SecretKeySpec(decryptedKey, decryptionMaterials.getAlgorithm().getDataKeyAlgo()));
+                    traceOnDecrypt(decryptionMaterials.getKeyringTrace());
+                    return;
+                } catch (Exception e) {
+                    LOGGER.info("Could not decrypt key due to: " + e.getMessage());
+                }
+            }
+        }
+
+        LOGGER.warning("Could not decrypt any data keys");
```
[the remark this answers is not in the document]
WesleyRosenblum: >In the case that onDecrypt gets an empty EDK list the correct behavior is to return. We should not log this warning in that case.
Good catch 

>I am also unsure if we want to log a warning here generally. What is the current behavior with the MK/MKP setup?
The Java ESDK basically has no logging anywhere. The current behavior with MK/MKP is to throw a CannotUnwrapDataKeyException if no key can be decrypted, but exception throwing is not specified in the spec so I didn't replicate that.

src/main/java/com/amazonaws/encryptionsdk/model/DecryptionMaterials.java
WesleyRosenblum: Actually I'm going to hold off on renaming this now, since that will break existing clients that have written their own CryptoMaterialsManager. Also I will only add the validation to the new setCleartextDataKey methods

WesleyRosenblum: Discussed with @seebees, I'm going to try defining new versions of DecryptionMaterials and EncryptionMaterials for exclusive use in Keyrings

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyringTest.java
WesleyRosenblum: I've added tests that check that the provider ID is as expected and the provider Information has the keyname as a prefix. I don't want to validate beyond that though, since the JceKeyCipher is separately tested and I want that to remain independent.

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyringTest.java
WesleyRosenblum: I do already check that the keyring trace has only 1 entry, and that keyring trace has only 2 flags and those flags are ENCRYPTED_DATA_KEY and SIGNED_ENCRYPTION_CONTEXT

src/test/java/com/amazonaws/encryptionsdk/keyrings/RawKeyringTest.java
WesleyRosenblum: Added

src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java
WesleyRosenblum: This class is going to revert to the original now that I am using an independent EncryptionMaterials and DecryptionMaterials

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
```diff
@@ -0,0 +1,189 @@
...
+        return keyringTrace;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        DecryptionMaterials that = (DecryptionMaterials) o;
+        return algorithm == that.algorithm &&
+                Objects.equals(plaintextDataKey, that.plaintextDataKey) &&
```
seebees: Is this constant time?  In the JS ESDK, we touch the plaintextDataKey in a similar way.  And it is also true that given the interface it does not _really_ matter.  However, we still went to the trouble of making the function constant time to avoid cognitive load on security engineers.

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
```diff
@@ -0,0 +1,189 @@
...
+                    String.format("Incorrect key length. Expected %s but got %s",
+                            algorithm.getDataKeyLength(), plaintextDataKey.getEncoded().length));
+            isTrue(algorithm.getDataKeyAlgo().equalsIgnoreCase(plaintextDataKey.getAlgorithm()),
+                    String.format("Incorrect key algorithm. Expected %s but got %s",
+                            algorithm.getDataKeyAlgo(), plaintextDataKey.getAlgorithm()));
+        }
+    }
+
+    /**
+     * Validates that a verification key is specified only if and only if
```
seebees:
```suggestion
     * Validates that a verification key is specified if and only if
```

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+        notNull(plaintextDataKey, "plaintextDataKey is required");
+        notNull(keyringTraceEntry, "keyringTraceEntry is required");
+        validatePlaintextDataKey(algorithm, plaintextDataKey);
+        this.plaintextDataKey = plaintextDataKey;
+        keyringTrace.add(keyringTraceEntry);
+    }
+
+    public PublicKey getVerificationKey() {
+        return verificationKey;
```
seebees: If `verificationKey` is `null` because the `algorithm` is not signed should this throw an error or just return `null`?

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * The algorithm to use for this decryption operation.
+     */
+    public CryptoAlgorithm getAlgorithm() {
+        return algorithm;
+    }
+
+    public SecretKey getPlaintextDataKey() {
+        return plaintextDataKey;
```
seebees: If the `plaintextDataKey` is not yet set, is it an error to request it?  Or is it just `null`?

src/main/java/com/amazonaws/encryptionsdk/keyrings/EncryptionMaterials.java
```diff
@@ -0,0 +1,237 @@
...
+     * Add an encrypted data key to the list of encrypted data keys.
+     *
+     * @param encryptedDataKey  The encrypted data key to add.
+     * @param keyringTraceEntry The keyring trace entry recording this action.
+     */
+    public void addEncryptedDataKey(EncryptedDataKey encryptedDataKey, KeyringTraceEntry keyringTraceEntry) {
+        notNull(encryptedDataKey, "encryptedDataKey is required");
+        notNull(keyringTraceEntry, "keyringTraceEntry is required");
+        encryptedDataKeys.add(encryptedDataKey);
+        keyringTrace.add(keyringTraceEntry);
```
seebees: I think that there should be some verification that the keyring trace is _valid_ . i.e. the flag `DECRYPTED_DATA_KEY ` is not valid when adding an encrypted data key :(

src/main/java/com/amazonaws/encryptionsdk/keyrings/EncryptionMaterials.java
```diff
@@ -0,0 +1,237 @@
...
+     */
+    public void setPlaintextDataKey(SecretKey plaintextDataKey, KeyringTraceEntry keyringTraceEntry) {
+        if (this.plaintextDataKey != null) {
+            throw new IllegalStateException("plaintextDataKey was already populated");
+        }
+        notNull(plaintextDataKey, "plaintextDataKey is required");
+        notNull(keyringTraceEntry, "keyringTraceEntry is required");
+        validatePlaintextDataKey(algorithm, plaintextDataKey);
+        this.plaintextDataKey = plaintextDataKey;
+        keyringTrace.add(keyringTraceEntry);
```
seebees: Similarly, the trace _must_ contain generate.

src/main/java/com/amazonaws/encryptionsdk/keyrings/EncryptionMaterials.java
```diff
@@ -0,0 +1,237 @@
...
+        notNull(keyringTraceEntry, "keyringTraceEntry is required");
+        encryptedDataKeys.add(encryptedDataKey);
+        keyringTrace.add(keyringTraceEntry);
+    }
+
+    /**
+     * A data key to be used as input for encryption.
+     */
+    public SecretKey getPlaintextDataKey() {
+        return plaintextDataKey;
```
seebees: This may not be as important for decrypt, but trying to get the plaintext data key before it is set is really bad.

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
WesleyRosenblum: SecretKeySpec relies on MessageDigest.isEqual for byte array comparison, which has been timing safe since Java 6 Update 17: https://www.oracle.com/technetwork/java/javase/6u17-141447.html

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
WesleyRosenblum: Throwing an error would likely reduce incorrect usage of this class and EncryptionMaterials, so I've gone ahead and made the optional plaintextDataKey, verificationKey, and signingKey throw IllegalStateException if they are not populated. I've also added has* methods so that these properties can be checked if they are populated.

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
WesleyRosenblum: Throws an IllegalStateException now

src/main/java/com/amazonaws/encryptionsdk/keyrings/EncryptionMaterials.java
WesleyRosenblum: Throws an IllegalStateException now

src/main/java/com/amazonaws/encryptionsdk/keyrings/EncryptionMaterials.java
WesleyRosenblum: I would prefer to leave that out of these classes. A user could make their own keyring implementations and decide they don't care about keyring traces, so they shouldn't need to populate it with some flags they don't necessarily care about.

src/main/java/com/amazonaws/encryptionsdk/keyrings/StandardKeyrings.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private StandardKeyrings() {
+    }
+
+    /**
+     * Constructs a {@code Keyring} which does local AES-GCM encryption
+     * decryption of data keys using the provided wrapping key.
+     *
+     * @param keyNamespace A UTF-8 encoded value that, together with the key name, identifies the wrapping key.
+     * @param keyName      A UTF-8 encoded value that, together with the key namespace, identifies the wrapping key.
```
robin-aws: Nit-pick: these are Java Strings, which means they are internally represented in UTF-16. And that's not really relevant anyway because the underlying `JceKeyCipher` class will call `keyName.getBytes(KEY_NAME_ENCODING)` (where `KEY_NAME_ENCODING == "UTF-8"`) and pass on the `keyNamespace` as a `String`. "A UTF-8 encoded value" really only makes sense if you're describing a `byte[]`.

I do realize the `JceKeyCipher` class has the same misleading comments, so perhaps you can punt fixing this for a separate PR for consistency.

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import static org.apache.commons.lang3.Validate.isTrue;
+import static org.apache.commons.lang3.Validate.notNull;
+
+/**
+ * Contains the cryptographic materials needed for a decryption operation with Keyrings.
+ */
+public final class DecryptionMaterials {
```
robin-aws: I worry a bit that having two Encryption/DecryptionMaterials classes in different packages might be a little confusing. The specification assumes they are the same data structure, and using the same name for different structures there would be REALLY confusing and/or verbose.

My specification PR (https://github.com/awslabs/aws-encryption-sdk-specification/pull/56) deals with part of this issue by introducing a `Data Key Materials` structure that's used at this level.

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
```diff
@@ -0,0 +1,227 @@
...
+
+import javax.crypto.SecretKey;
+import java.security.PublicKey;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+import static org.apache.commons.lang3.Validate.isTrue;
+import static org.apache.commons.lang3.Validate.notNull;
```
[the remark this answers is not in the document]
WesleyRosenblum: I've been using Validate.notNull since forever so I hadn't seen requireNonNull before :-) I'll switch it. 

Actually I'd like to start using Lombok to handle this stuff with annotations, thoughts on that?

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
```diff
@@ -0,0 +1,227 @@
...
+
+        private Builder(DecryptionMaterials result) {
+            this.algorithmSuite = result.algorithmSuite;
+            this.plaintextDataKey = result.plaintextDataKey;
+            this.verificationKey = result.verificationKey;
+            this.encryptionContext = result.encryptionContext;
+            this.keyringTrace = result.keyringTrace;
+        }
+
+        public Builder setAlgorithmSuite(CryptoAlgorithm algorithmSuite) {
```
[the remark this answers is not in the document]
WesleyRosenblum: That's a good example to follow, I'll make the change

src/main/java/com/amazonaws/encryptionsdk/keyrings/EncryptionMaterials.java
```diff
@@ -0,0 +1,270 @@
...
+            return new EncryptionMaterials(this);
+        }
+
+        public Builder setAlgorithmSuite(CryptoAlgorithm algorithmSuite) {
+            this.algorithmSuite = algorithmSuite;
+            return this;
+        }
+
+        public Builder setEncryptionContext(Map<String, String> encryptionContext) {
+            this.encryptionContext = Collections.unmodifiableMap(new HashMap<>(encryptionContext));
```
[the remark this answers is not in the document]
WesleyRosenblum: It is making a copy 

```
new HashMap<>(encryptionContext)

File: src/main/java/com/amazonaws/encryptionsdk/keyrings/Keyring.java [hunk 26]
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.List;
+
+/**
+ * Keyrings are responsible for the generation, encryption, and decryption of data keys.
+ */
+public interface Keyring {
+
+    /**
+     * Generate a data key if not present and encrypt it using any available wrapping key
```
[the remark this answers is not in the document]
WesleyRosenblum: Changed it to be more general

src/main/java/com/amazonaws/encryptionsdk/keyrings/StandardKeyrings.java
WesleyRosenblum: Yeah I took that from the RawAesKeyring spec. Should it change there too? 

I'll fix it here for now

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
WesleyRosenblum: I'll leave this for now and we can revisit it after getting some clarity on the spec. I originally had it as the same data structure but there were issues with backwards compatibility that made that undesirable.

src/main/java/com/amazonaws/encryptionsdk/keyrings/StandardKeyrings.java
lavaleri: per updating the spec: I think that an argument can be made that using Java strings and then erroring if there is an issue converting the string to a UTF-8 encoded byte[] later is still following the behavior defined by the specification.

However, there is the issue that in the Java implementation it is possible to initialize keyrings which are unable to produce EDKs (at least for our "standard" keyrings). Is it considered being out of line with the spec to call something a "keyring" if it is unable to perform the behaviors of a keyring? Or is the fact that such a keyring only results in errors mean it's still "in line" with the spec (following the philosophy that: if you can't follow the spec behavior, you MUST fail). For now it is my opinion that we go with the latter approach. Thoughts?

robin-aws: Interesting point - the specification should probably say "Unicode string" instead of "UTF-8 encoded value". It would be silly to use byte arrays everywhere instead of the language-native String type.

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
robin-aws: I definitely prefer using two different data structures, but if you're going ahead with this I would also cut an issue to track the deviation from the spec - it won't be consistent with either the current version or my PR.

src/main/java/com/amazonaws/encryptionsdk/keyrings/EncryptionMaterials.java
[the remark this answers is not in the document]
robin-aws: Doh, reading comprehension fail. :)

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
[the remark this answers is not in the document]
robin-aws: I've been very wary about using Lombok in our open source projects, since it introduces another dependency. There are known interaction issues between Lombok and Kotlin too, and given the choice I'd prefer to support Kotlin better. :)

src/main/java/com/amazonaws/encryptionsdk/keyrings/DecryptionMaterials.java
WesleyRosenblum: https://github.com/aws/aws-encryption-sdk-java/issues/146

Issue #143, opened 2019-11-18
Published Javadoc is out of date.


lizroth:
https://aws.github.io/aws-encryption-sdk-java/javadoc/com/amazonaws/encryptionsdk/AwsCrypto.html#encryptString(com.amazonaws.encryptionsdk.CryptoMaterialsManager,java.lang.String) is deprecated and has had a Javadoc update, but https://aws.github.io/aws-encryption-sdk-java/javadoc/ is still showing version 1.6.0's Javadoc.

At the least, Javadoc needs a manual update. Getting it into the release process as well would be good, but there might be more blockers to getting that done right away.

Possibly related to https://github.com/aws/aws-encryption-sdk-java/issues/90 and https://github.com/aws/aws-encryption-sdk-java/issues/91

PR #144, opened 2019-11-18
update javadoc to 1.6.1

mattsb42-aws:
fixes: #143 

*Description of changes:*
Updating with contents of javadoc jar in 1.6.1 release artifacts.

# Check any applicable:


mattsb42-aws:
that linecount diff does not sound right... :/

SalusaSecondus:
This is reasonable. Different versions of the `javadoc` binary can produce very different looking output.

PR #148, opened 2019-12-12
Define the MultiKeyring

WesleyRosenblum:
Description of changes:

This change contains the initial implementation of the Multi Keyring.

# Check any applicable:

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
@@ -36,7 +36,7 @@ class MultiKeyring implements Keyring {
 
     MultiKeyring(Keyring generatorKeyring, List<Keyring> childrenKeyrings) {
         this.generatorKeyring = generatorKeyring;
-        this.childrenKeyrings = childrenKeyrings == null ? emptyList() : unmodifiableList(childrenKeyrings);
+        this.childrenKeyrings = childrenKeyrings == null ? emptyList() : unmodifiableList(new ArrayList<>(childrenKeyrings));
 
         isTrue(this.generatorKeyring != null || !this.childrenKeyrings.isEmpty(),
                 "At least a generator keyring or children keyrings must be defined");
--- a/src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except
+ * in compliance with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
```diff
@@ -0,0 +1,102 @@
...
+ * modify the encryption or decryption materials using more than one keyring.
+ */
+class MultiKeyring implements Keyring {
+
+    final Keyring generatorKeyring;
+    final List<Keyring> childrenKeyrings;
+
+    MultiKeyring(Keyring generatorKeyring, List<Keyring> childrenKeyrings) {
+        this.generatorKeyring = generatorKeyring;
+        this.childrenKeyrings = childrenKeyrings == null ? emptyList() : unmodifiableList(childrenKeyrings);
```
SalusaSecondus: Might be worth making a defensive copy of `childrenKeyrings`. `unmodifiableList` keeps you from modifying it, not anyone else.

src/main/java/com/amazonaws/encryptionsdk/keyrings/StandardKeyrings.java
```diff
@@ -53,4 +54,17 @@ public static Keyring rawAes(String keyNamespace, String keyName, SecretKey wrap
...
+    /**
+     * Constructs a {@code Keyring} which combines other keyrings, allowing one OnEncrypt or OnDecrypt call
+     * to modify the encryption or decryption materials using more than one keyring.
+     *
+     * @param generatorKeyring A keyring that can generate data keys. Required if childrenKeyrings is empty.
+     * @param childrenKeyrings A list of keyrings to be used to modify the encryption or decryption materials.
+     *                         Required if generatorKeyring is null.
+     * @return The {@link Keyring}
+     */
+    public static Keyring multi(Keyring generatorKeyring, List<Keyring> childrenKeyrings) {
```
SalusaSecondus: NitPick: It might be worth adding this method also:

```
public static multi(Keyring generatorKeyring, Keyring ... childrenKeyRings) {
   return multi(generatorKeyring, Arrays.toList(childrenKeyRings));
}
```

src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
WesleyRosenblum: Done. I'll make this change in the KmsKeyring too

src/main/java/com/amazonaws/encryptionsdk/keyrings/StandardKeyrings.java
WesleyRosenblum: Added

src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public void onDecrypt(DecryptionMaterials decryptionMaterials, List<? extends EncryptedDataKey> encryptedDataKeys) {
+        requireNonNull(decryptionMaterials, "decryptionMaterials are required");
+        requireNonNull(encryptedDataKeys, "encryptedDataKeys are required");
+
+        if (decryptionMaterials.hasPlaintextDataKey()) {
+            return;
+        }
```
seebees: These are all important for every keyring.
The fact that this code is here in the MultiKeyring implies to me that it will need to be in every keyring?
While not the end of the world, 
this does mean that if we have people write their own keyring
they will need to remember to include this.

Issue #150, opened 2020-01-13
Master Key to Keyring API Evolution


WesleyRosenblum:
### Introduction
The Java version of the AWS Encryption SDK, having been developed prior to all other language variants of the AWS Encryption SDK, contains configuration and API constructs that do not exist in the more recently developed versions. Namely, the Master Key/Master Key Provider (MKP) construct has been replaced by Keyrings in the C, Javascript, and all future versions of the ESDK. To bring the Java ESDK in line with the other ESDKs, Keyring support is currently being added (see #102 ). However, since Master Key Providers are present in both the input and output of the Java ESDK, the process of introducing Keyrings while minimizing disruption to existing customers is complicated. This issue summarizes the current state of Master Key Providers in the Java ESDK and raises the following questions regarding the evolution of the Java ESDK API:

* Should we provide a Master Key Provider Keyring to ease transitioning to Keyrings?
* How should the result object of encryption and decryption operations be updated to support Keyrings?
* Should a new major version be introduced?

### Current State
The AwsCrypto class serves as the primary entry-point to the Java ESDK. This class contains 30 public methods interacting with Master Key Providers, which are overloaded variations of the following 4 methods: 

* estimateCiphertextSize, for estimating the output length of encrypting a plaintext
* encryptData/decryptData, the primary encryption and decryption methods
* createEncryptingStream/createDecryptingStream, for encrypting/decrypting input streams
* ~~encryptString~~/~~decryptString~~, deprecated methods for encrypting/decrypting strings

Each of these methods use either a MasterKeyProvider or a CryptoMaterialsManager to perform its encryption/decryption objective. When a MasterKeyProvider is specified, the given MKP is wrapped in the DefaultCryptoMaterialsManager and handed off to the corresponding CryptoMaterialsManager based method. All of these methods (save for estimateCiphertextSize which returns a long) return a CryptoResult or a CryptoInputStream/CryptoOutputStream that wraps a CryptoResult. CryptoResult is a generic type that is parameterized with the MasterKey type that was specified to the method. CryptoResult contains a getMasterKeys method that in the case of encryption returns all the MasterKeys used to encrypt the ciphertext, and in the case of decryption returns the single MasterKey used to decrypt the ciphertext. 

### Should we provide a Master Key Provider Keyring?

[...]

PR #151, opened 2020-01-17
Incorporate Keyrings into AwsCrypto and deprecate MasterKeyProviders.

WesleyRosenblum:
Description of changes:

This change incorporates Keyrings into AwsCrypto and deprecates MasterKeys and MasterKeyProviders.

See #150 for a discussion on the evolution of the API as we transition to Keyrings
See #153 for a discussion on the ergonomics of Keyrings

# Check any applicable:


src/examples/java/com/amazonaws/crypto/examples/FileStreamingExample.java
```diff
@@ -40,55 +43,83 @@
...
+            encryptingStream.close();
+        }
 
-        FileOutputStream out = new FileOutputStream(srcFile + ".encrypted");
-        IOUtils.copy(encryptingStream, out);
-        encryptingStream.close();
-        out.close();
+        // 8. Create the decrypting stream.
+        final AwsCryptoInputStream decryptingStream =
+                crypto.createDecryptingStream(config, new FileInputStream(encryptedFile));
```
robin-aws: I was about to comment that this should use a try-with-resources too, but then remembered that only works if you're directly instantiating the object :(

In any case, you'll want a manual try/finally to make sure this is closed as well.

WesleyRosenblum: I tested it out with try-with-resources and it seems to auto close even without directly instantiating the AwsCryptoInputStream

robin-aws: Ah, now I recall that you just don't get the IDE warning unless you're instantiating the type directly inside the try (...).

WesleyRosenblum: I've updated to use try-with-resources for the AwsCryptoInputStream and AwsCryptoOutputStream. I realize I could put both the AwsCrypto streams and the FiltOutputStreams into the same try-with-resources statement like:

``` java
try(final AwsCryptoInputStream decryptingStream = crypto.createDecryptingStream(config, new FileInputStream(encryptedFile));
           final FileOutputStream out = new FileOutputStream(decryptedFile)) {
```

, but it didn't flow as well for me in terms of delineating the steps, so I separated them. Let me know if you'd prefer them together though

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -383,8 +474,12 @@ public long estimateCiphertextSize(
 
     /**
      * @see #decryptData(MasterKeyProvider, byte[])
+     *
+     * @deprecated {@link MasterKeyProvider}s have been deprecated in favor of {@link Keyring}s.
+     *              Replaced by {@link #decryptData(AwsCryptoConfig, byte[])}
```
SalusaSecondus: There is value is still having a decrypt method which takes a `ParsedCiphertext` as an argument. That lets a caller inspect the ciphertext prior to decryption without incurring a double parsing cost.

WesleyRosenblum: Where does a caller have access to a ParsedCiphertext without creating it themselves? The AwsCryptoResult has CiphertextHeaders, but no ParsedCiphertext

SalusaSecondus: They would create it themselves. That is allowed. Then they can inspect the headers prior to decryption (though they have to remember that the data is untrusted at that point).

WesleyRosenblum: Got it, I've made decryptData( ParsedCipherText) public

src/examples/java/com/amazonaws/crypto/examples/BasicEncryptionExample.java
```diff
@@ -48,32 +52,44 @@ static void encryptAndDecrypt(final String keyArn) {
...
 
-        // 5. Decrypt the data
-        final CryptoResult<byte[], KmsMasterKey> decryptResult = crypto.decryptData(masterKeyProvider, ciphertext);
+        // 7. Decrypt the data
+        final AwsCryptoResult<byte[]> decryptResult = crypto.decryptData(config, ciphertext);
 
-        // 6. Before verifying the plaintext, verify that the customer master key that
-        // was used in the encryption operation was the one supplied to the master key provider.
-        if (!decryptResult.getMasterKeyIds().get(0).equals(keyArn)) {
+        // 8. Before verifying the plaintext, verify that the key that was used in the encryption
```
juneb: Before returning the plaintext, verify that the same CMK was used to encrypt and decrypt the data. (This is kind of silly in symmetric encryption.)

Better: Before returning the plaintext, verify that the CMK that was used to decrypt the encrypted data key was the CMK in the encryption keyring. (Again, I can't imagine how this can fail unless both keyrings are discovery keyrings.)

Actually, the more interesting test of our keyring implementation is whether each encrypted data key was encrypted by a CMK specified in the keyring. Then, unless you specify a discovery keyring, you can verify that the CMK that decrypted the encrypted data key was one of the CMKs in the decryption keyring. 

In all of these cases, a failure means that the ESDK failed; not that the user did anything wrong.

src/examples/java/com/amazonaws/crypto/examples/BasicEncryptionExample.java
```diff
@@ -48,32 +52,44 @@ static void encryptAndDecrypt(final String keyArn) {
         // 1. Instantiate the SDK
         final AwsCrypto crypto = new AwsCrypto();
```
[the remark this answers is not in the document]
WesleyRosenblum: AwsCrypto is still the name of the class containing the main entry points to the SDK. AwsCryptoConfig is the name of the input to the new methods in this class

src/examples/java/com/amazonaws/crypto/examples/BasicEncryptionExample.java
```diff
@@ -48,32 +52,44 @@ static void encryptAndDecrypt(final String keyArn) {
...
+                .encryptionContext(encryptionContext)
+                .build();
+
+        // 6. Encrypt the data
+        final AwsCryptoResult<byte[]> encryptResult = crypto.encryptData(config, EXAMPLE_DATA);
         final byte[] ciphertext = encryptResult.getResult();
 
-        // 5. Decrypt the data
-        final CryptoResult<byte[], KmsMasterKey> decryptResult = crypto.decryptData(masterKeyProvider, ciphertext);
+        // 7. Decrypt the data
```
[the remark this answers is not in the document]
WesleyRosenblum: I've added a note

src/examples/java/com/amazonaws/crypto/examples/BasicEncryptionExample.java
```diff
@@ -48,32 +52,44 @@ static void encryptAndDecrypt(final String keyArn) {
...
-        // 2. Instantiate a KMS master key provider
-        final KmsMasterKeyProvider masterKeyProvider = KmsMasterKeyProvider.builder().withKeysForEncryption(keyArn).build();
+        // 2. Instantiate a KMS Client Supplier. This example uses the default client supplier but you can
+        //    also configure the credentials provider, client configuration and other settings as necessary
+        final KmsClientSupplier clientSupplier = KmsClientSupplier.builder().build();
+
+        // 3. Instantiate a KMS Keyring, supplying the keyArn as the generator for generating a data key.
+        //    For this example, empty lists are provided for grant tokens and additional keys to encrypt the data
+        //    key with, but those can be supplied as necessary.
+        final Keyring keyring = StandardKeyrings.kms(clientSupplier, emptyList(), emptyList(), keyArn);
```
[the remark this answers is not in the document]
WesleyRosenblum: I've adjusted the comment to mention CMK alias and alias ARN

src/examples/java/com/amazonaws/crypto/examples/BasicEncryptionExample.java
WesleyRosenblum: I've reworded this to match your suggestion. I would say that the example code isn't really meant as a test, this verification is really just to introduce the Keyring Trace so the user knows how to get the information that used to be in decryptResult.getMasterKeyIds

src/examples/java/com/amazonaws/crypto/examples/EscrowedEncryptExample.java
```diff
@@ -1,168 +1,172 @@
...
- * 
+ *
  * The sample encrypts data under both a KMS customer master key (CMK) and an "escrowed" RSA key pair
- * so that either key alone can decrypt it. You might commonly use the KMS CMK for decryption. However, 
+ * so that either key alone can decrypt it. You might commonly use the KMS CMK for decryption. However,
  * at any time, you can use the private RSA key to decrypt the ciphertext independent of KMS.
  *
- * This sample uses the JCEMasterKey class to generate a RSA public-private key pair
- * and saves the key pair in memory. In practice, you would store the private key in a secure offline 
+ * This sample uses an RawRsaKeyring to generate a RSA public-private key pair
```
mattsb42-aws:
```suggestion
 * This sample uses a RawRsaKeyring to generate a RSA public-private key pair
```

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -270,14 +314,54 @@ public long estimateCiphertextSize(
...
+    /**
+     * Returns an encrypted form of {@code plaintext} that has been protected with either the {@link CryptoMaterialsManager}
+     * or the {@link Keyring} specified in the {@link AwsCryptoConfig}.
+     *
+     * @param config The AwsCryptoConfig containing either a {@link CryptoMaterialsManager} or a {@link Keyring} and
+     *               optionally an EncryptionContext
+     * @param plaintext The plaintext to encrypt
+     * @return An {@link AwsCryptoResult} containing the encrypted data
+     */
+    public AwsCryptoResult<byte[]> encryptData(final AwsCryptoConfig config,
```
mattsb42-aws: Since we're fundamentally changing the footprint of this API, I would love to take this opportunity to rename these to simply `encrypt` and `decrypt`. This is something I've been wanting to do for a long time but never had a justification. Doing this now has the benefit not only of simplifying the API but also making it explicitly clear what calls are impacted by that change not only by secondary identifiers like the return and parameter types, but by the method name itself.

src/main/java/com/amazonaws/encryptionsdk/EncryptRequest.java
```diff
@@ -0,0 +1,94 @@
...
+        this.encryptionContext = builder.encryptionContext;
+    }
+
+    public byte[] plaintext() {
+        return this.plaintext;
+    }
+
+    public Map<String, String> encryptionContext() {
+        return this.encryptionContext;
+    }
```
praus: NP: I think it's the right call not to make defensive copies in this class. I would recommend putting a comment explaining that it's intentional. If FindBugs are used (I can't recall if they are or not) you'll also need to suppress warnings.

WesleyRosenblum: Actually there is a defensive copy of encryptionContext created in the Builder. Should I take that out?

mattsb42-aws:
Not entirely relevant to this PR, but since we don't have a better place and this is where you said you'll be doing the change:

As mentioned here[1], can we make sure that we call the AWS KMS keyring the "AWS KMS" keyring, not just the "KMS" keyring?

[1] https://github.com/awslabs/aws-encryption-sdk-specification/issues/67

src/main/java/com/amazonaws/encryptionsdk/EncryptRequest.java
WesleyRosenblum: I've added a comment for plaintext and ciphertext

src/test/java/com/amazonaws/encryptionsdk/internal/StaticKeyring.java
```diff
@@ -0,0 +1,189 @@
...
+import java.security.spec.X509EncodedKeySpec;
+import java.util.List;
+
+import static com.amazonaws.encryptionsdk.EncryptedDataKey.PROVIDER_ENCODING;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Static implementation of the {@link Keyring} interface that should only
+ * used for unit-tests.
+ * <p>
```
mattsb42-aws: This needs a different name. "Static" is the name we have been reserving for the successor to "raw".

WesleyRosenblum: Renamed to TestKeyring

src/examples/java/com/amazonaws/crypto/examples/BasicEncryptionExample.java
```diff
@@ -39,44 +42,52 @@
...
         final Map<String, String> encryptionContext = Collections.singletonMap("ExampleContextKey", "ExampleContextValue");
 
-        // 4. Encrypt the data
-        final CryptoResult<byte[], KmsMasterKey> encryptResult = crypto.encryptData(masterKeyProvider, EXAMPLE_DATA, encryptionContext);
+        // 4. Encrypt the data with the keyring and encryption context
+        final AwsCryptoResult<byte[]> encryptResult = crypto.encrypt(
+                EncryptRequest.builder()
+                    .keyring(keyring)
+                    .encryptionContext(encryptionContext)
+                    .plaintext(EXAMPLE_DATA).build());
```
robin-aws: In all the API methods that have request objects, you could support the AWS SDK Java v2 idiom of a lambda parameter instead of a user-managed builder object:

```java
crypto.encrypt(request -> 
        request.keyring(keyring)
               .ciphertext(ciphertext)
               .plaintext(EXAMPLE_DATA));
```

Check the SDK docs for formatting and naming idioms.

src/examples/java/com/amazonaws/crypto/examples/EscrowedEncryptExample.java
```diff
@@ -1,168 +1,157 @@
...
-
-        IOUtils.copy(in, encryptingStream);
-        in.close();
-        encryptingStream.close();
+        // 2. Instantiate an AWS KMS Keyring, supplying the keyArn as the generator for generating a data key.
+        final Keyring kmsKeyring = StandardKeyrings.awsKms(kmsArn);
+
+        // 3. Instantiate a RawRsaKeyring
+        //    Because the user does not have access to the private escrow key,
+        //    they pass in "null" for the private key parameter.
```
robin-aws:
```suggestion
        //    they do not provide the private key parameter.
```

src/examples/java/com/amazonaws/crypto/examples/FileStreamingExample.java
```diff
@@ -40,55 +45,83 @@
...
+            }
         }
 
-        // Return the plaintext data
-        out = new FileOutputStream(srcFile + ".decrypted");
-        IOUtils.copy(decryptingStream, out);
-        decryptingStream.close();
-        out.close();
+        // 7. Create the decrypting input stream with the keyring.
+        try(final AwsCryptoInputStream decryptingStream = crypto.createDecryptingInputStream(
```
robin-aws:
```suggestion
        try (final AwsCryptoInputStream decryptingStream = crypto.createDecryptingInputStream(
```

src/examples/java/com/amazonaws/crypto/examples/FileStreamingExample.java
```diff
@@ -97,4 +130,21 @@ private static SecretKey retrieveEncryptionKey() {
...
+        try (BufferedReader file1Reader = Files.newBufferedReader(file1.toPath());
+             BufferedReader file2Reader = Files.newBufferedReader(file2.toPath())) {
+            String file1Line;
+            String file2Line;
+
+            while ((file1Line = file1Reader.readLine()) != null &&
+                    (file2Line = file2Reader.readLine()) != null) {
+                assert Objects.equals(file1Line, file2Line);
+            }
+
```
robin-aws:
```suggestion
```

src/examples/java/com/amazonaws/crypto/examples/BasicEncryptionExample.java
WesleyRosenblum: Thanks for the tip, I've added these convenience methods

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -27,31 +27,46 @@
...
  * {@link AwsCrypto} provides the primary way to encrypt/decrypt data. It can operate on
- * byte-arrays, streams, or {@link java.lang.String Strings}. This data is encrypted using the
- * specifed {@link CryptoAlgorithm} and a {@link DataKey} which is unique to each encrypted message.
- * This {@code DataKey} is then encrypted using one (or more) {@link MasterKey MasterKeys}. The
- * process is reversed on decryption with the code selecting a copy of the {@code DataKey} protected
- * by a usable {@code MasterKey}, decrypting the {@code DataKey}, and then decrypted the message.
+ * byte-arrays and streams. This data is encrypted using the specified {@link CryptoAlgorithm} and a
+ * {@code plaintextDataKey} which is unique to each encrypted message. This {@code plaintextDataKey} is then encrypted
+ * using one (or more) {@link MasterKey MasterKeys} or a {@link Keyring}. The process is reversed on decryption with the
+ * code selecting a copy of the {@code EncryptedDataKey} protected by a usable {@code MasterKey} or {@code Keyring},
```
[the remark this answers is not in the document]
WesleyRosenblum: It doesn't mean much here, I'll clean up the wording

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -162,34 +181,31 @@ public int getEncryptionFrameSize() {
     /**
      * Returns the best estimate for the output length of encrypting a plaintext with the provided
      * {@code plaintextSize} and {@code encryptionContext}. The actual ciphertext may be shorter.
+     *
+     * @deprecated Replaced by {@link #estimateCiphertextSize(EstimateCiphertextSizeRequest)}
```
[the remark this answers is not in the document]
WesleyRosenblum: I would prefer to deprecate them for consistency and for the flexibility it gives us if we want to add further parameters later on. And if we kept `estimateCiphertextSize(final CryptoMaterialsManager materialsManager, final int plaintextSize)` we'd probably also keep  `estimateCiphertextSize(final CryptoMaterialsManager materialsManager, final int plaintextSize, final Map<String, String> encryptionContext)` and you get the same issue with too many overloaded methods.

src/main/java/com/amazonaws/encryptionsdk/internal/MessageCryptoHandler.java
```diff
@@ -49,5 +51,14 @@
      * {@link MasterKey}s used to protect the data. In the decryption flow, it is the single
      * {@link MasterKey} actually used to decrypt the data.
      */
+    @Deprecated
```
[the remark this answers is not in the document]
WesleyRosenblum: added

src/main/java/com/amazonaws/encryptionsdk/AwsCryptoResult.java
```diff
@@ -0,0 +1,100 @@
...
+     * The encryption context.
+     *
+     * @return The encryption context
+     */
+    public Map<String, String> getEncryptionContext() {
+        return encryptionContext;
+    }
+
+    /**
+     * Convenience method equivalent to {@link #getHeaders()}.{@code getCryptoAlgoId()}.
```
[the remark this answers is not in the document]
WesleyRosenblum: Fixed

src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java
```diff
@@ -536,9 +539,18 @@ public CiphertextHeaders getHeaders() {
 
     @Override
     public List<K> getMasterKeys() {
+        if(dataKey_.getMasterKey() == null) {
```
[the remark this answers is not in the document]
WesleyRosenblum: When a keyring is used, `dataKey_.getMasterKey()` will return null, so I wanted the returned list to be an empty list, rather than a list of one null element. It looks like I forgot to deprecate this method though (and a couple others in this class), I'll fix that

src/main/java/com/amazonaws/encryptionsdk/caching/CachingCryptoMaterialsManager.java
```diff
@@ -100,13 +101,33 @@ public Builder withBackingMaterialsManager(CryptoMaterialsManager backingCMM) {
...
         public Builder withMasterKeyProvider(MasterKeyProvider mkp) {
             return withBackingMaterialsManager(new DefaultCryptoMaterialsManager(mkp));
         }
 
+        /**
+         * Sets the {@link Keyring} that should be queried when the {@link CachingCryptoMaterialsManager}
+         * incurs a cache miss.
+         *
+         * You can set either a Keyring or a CryptoMaterialsManager to back the CCMM - the last value set will
+         * be used.
```
robin-aws: This is a good callout. I didn't like it at first until I realized all builders are like this (silently overwriting existing parameters if you call them multiple times).

Can you specify that all three of the builder methods overwrite each other, on all three methods? That will make this sit better.

src/main/java/com/amazonaws/encryptionsdk/keyrings/StandardKeyrings.java
```diff
@@ -31,48 +28,74 @@ private StandardKeyrings() {
...
+    }
+
+    /**
+     * Returns a {@link AwsKmsKeyringBuilder} for use in constructing a keyring which interacts with
+     * AWS Key Management Service (KMS) to create, encrypt, and decrypt data keys using AWS KMS defined
+     * Customer Master Keys (CMKs).
+     *
+     * @return The {@link AwsKmsKeyringBuilder}
+     */
+    public static AwsKmsKeyringBuilder awsKms() {
```
robin-aws: Nit: The lack of symmetry in these three methods is a little odd, given one produces the keyring itself and the other two builders. Perhaps add "...Builder" to the latter two? And perhaps to the two raw methods above too in that case?

src/main/java/com/amazonaws/encryptionsdk/keyrings/AwsKmsKeyringBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+        return this;
+    }
+
+    /**
+     * Constructs the {@link Keyring} instance.
+     *
+     * @return The {@link Keyring} instance
+     */
+    public Keyring build() {
+        if(awsKmsClientSupplier == null) {
```
robin-aws:
```suggestion
        if (awsKmsClientSupplier == null) {
```

src/main/java/com/amazonaws/encryptionsdk/keyrings/AwsKmsKeyringBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+    private AwsKmsClientSupplier awsKmsClientSupplier;
+    private List<String> grantTokens;
+    private List<AwsKmsCmkId> keyIds;
+    private AwsKmsCmkId generatorKeyId;
+    private final boolean isDiscovery;
+
+    private AwsKmsKeyringBuilder(boolean isDiscovery) {
+        // Use AwsKmsKeyringBuilder.standard() or StandardKeyrings.awsKms() to instantiate
+        // or AwsKmsKeyringBuilder.discovery() or StandardKeyrings.awsKmsDiscovery() to instantiate
+        // a discovery keyring builder.
```
robin-aws: Wut :)

src/main/java/com/amazonaws/encryptionsdk/keyrings/KeyringTrace.java
```diff
@@ -65,4 +64,17 @@ public String toString() {
                 .append("entries", entries)
                 .toString();
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
robin-aws: Consider making this class final since it's a "value class"

src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
```diff
@@ -45,30 +45,34 @@
...
 
-        if (!encryptionMaterials.hasCleartextDataKey()) {
+        if (!resultMaterials.hasCleartextDataKey()) {
             throw new AwsCryptoException("Either a generator keyring must be supplied that produces a cleartext " +
                     "data key or a cleartext data key must already be present in the encryption materials.");
         }
 
         for (Keyring keyring : childrenKeyrings) {
-            keyring.onEncrypt(encryptionMaterials);
+            resultMaterials = keyring.onEncrypt(resultMaterials);
```
robin-aws: Not relevant to this particular PR, but are we clear on whether there are any "non-fatal" exceptions Keyrings can throw that would still allow for succeeding on a different child keyring?

src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
```diff
@@ -83,11 +87,11 @@ public void onDecrypt(DecryptionMaterials decryptionMaterials, List<? extends En
 
         for (Keyring keyring : keyringsToDecryptWith) {
             try {
-                keyring.onDecrypt(decryptionMaterials, encryptedDataKeys);
+                final DecryptionMaterials resultMaterials = keyring.onDecrypt(decryptionMaterials, encryptedDataKeys);
```
robin-aws: Same comment as with `onEncrypt`

src/main/java/com/amazonaws/encryptionsdk/keyrings/RawKeyring.java
```diff
@@ -65,53 +65,54 @@
...
                             new KeyringTraceEntry(keyNamespace, keyName, decryptTraceFlags()));
-                    return;
                 } catch (Exception e) {
                     LOGGER.info("Could not decrypt key due to: " + e.getMessage());
                 }
             }
         }
 
-        LOGGER.warning("Could not decrypt any data keys");
+        return decryptionMaterials;
```
robin-aws: +1 to getting rid of the log :)

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkId.java
```diff
@@ -0,0 +1,138 @@
...
+ * A representation of an AWS KMS Customer Master Key Identifier, which may be one either a
+ * key ID, key Amazon Resource Name (ARN), alias name, or alias ARN.
+ */
+public class AwsKmsCmkId {
+
+    private static final String ARN_PREFIX = "arn:";
+    private String keyId;
+
+    private AwsKmsCmkId(String keyId) throws MalformedArnException {
+        Validate.notBlank(keyId, "keyId must be a valid CMK Amazon Resource Name (ARN), CMK Alias, Key Id or Alias ARN");
```
robin-aws: I think you'd be a lot more helpful to state explicitly "keyId cannot be blank"

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkId.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Override
+    public String toString() {
+        return keyId;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
robin-aws: Consider making this class final since it's a "value class"

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkId.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.Validate;
+
+import java.util.Objects;
+
+/**
+ * A representation of an AWS KMS Customer Master Key Identifier, which may be one either a
+ * key ID, key Amazon Resource Name (ARN), alias name, or alias ARN.
+ */
+public class AwsKmsCmkId {
```
robin-aws: I like having this explicit type. Curious why we don't care about the "alias/" prefix any more though.

src/main/java/com/amazonaws/encryptionsdk/model/EncryptionMaterials.java
```diff
@@ -70,20 +71,28 @@ public CryptoAlgorithm getAlgorithm() {
...
         requireNonNull(encryptedDataKey, "encryptedDataKey is required");
         requireNonNull(keyringTraceEntry, "keyringTraceEntry is required");
+
+        final List<KeyBlob> encryptedDataKeys = new ArrayList<>(getEncryptedDataKeys());
         encryptedDataKeys.add(encryptedDataKey);
-        keyringTrace.add(keyringTraceEntry);
+
+        return toBuilder()
+                .setEncryptedDataKeys(encryptedDataKeys)
+                .setKeyringTrace(keyringTrace.with(keyringTraceEntry))
```
robin-aws: Nit: shouldn't these be "with" methods?

(Note I'm not 100% sure if the builder existed already, if so then don't change this)

src/main/java/com/amazonaws/encryptionsdk/CryptoInputStream.java
```diff
@@ -254,4 +257,18 @@ public void setMaxInputLength(long size) {
                 (List<K>) cryptoHandler_.getMasterKeys(),
                 cryptoHandler_.getHeaders());
     }
+
+    AwsCryptoResult<AwsCryptoInputStream> getAwsCryptoResult(AwsCryptoInputStream awsCryptoInputStream) throws IOException {
+        while (!cryptoHandler_.getHeaders().isComplete()) {
+            if (fillOutBytes() == -1) {
+                throw new BadCiphertextException("No CiphertextHeaders found.");
+            }
+        }
```
robin-aws: Where was this refactored from? Or is it new?

src/main/java/com/amazonaws/encryptionsdk/model/EncryptionMaterials.java
WesleyRosenblum: The builder already existed

src/main/java/com/amazonaws/encryptionsdk/CryptoInputStream.java
WesleyRosenblum: It is copied from CryptoInputStream.getCryptoResult()

src/main/java/com/amazonaws/encryptionsdk/CreateDecryptingInputStreamRequest.java
```diff
@@ -0,0 +1,79 @@
...
+        super(builder);
+
+        requireNonNull(builder.inputStream, "inputStream is required");
+        this.inputStream = builder.inputStream;
+    }
+
+    /**
+     * The {@link InputStream} to be read from.
+     *
+     * @return The {@link InputStream} to be read from.
```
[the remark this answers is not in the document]
WesleyRosenblum: No standards that I'm aware of, I was just trying to be complete at least on our public APIs. Looking at some examples in the JDK, it seems redundant javadocs are preferred over missing tags.

src/main/java/com/amazonaws/encryptionsdk/DefaultCryptoMaterialsManager.java
```diff
@@ -13,131 +15,185 @@
...
+        this.mkp = null;
+    }
 
-        CryptoAlgorithm algo = request.getRequestedAlgorithm();
-        if (algo == null) {
-            algo = AwsCrypto.getDefaultCryptoAlgorithm();
+    @Override
+    public EncryptionMaterials getMaterialsForEncrypt(EncryptionMaterialsRequest request) {
+        if(keyring != null) {
+            return getEncryptionMaterialsForKeyring(request);
```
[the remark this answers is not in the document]
WesleyRosenblum: Done, and I've done the same for decrypt materials

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkId.java
WesleyRosenblum: I'll go with "keyId is required" to match the rest of the validation code

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkId.java
WesleyRosenblum: Previously we cared about the alias/ prefix because if it wasn't an alias we would assume it was an ARN. We then decide to continue supporting "raw" key IDs, which could basically be anything, so that assumption no longer holds. So now I just check if it looks like an arn and thus will validate it, otherwise anything is allowed.

src/main/java/com/amazonaws/encryptionsdk/keyrings/AwsKmsKeyringBuilder.java
WesleyRosenblum: I've revised the wording a bit.

src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
WesleyRosenblum: The spec just says "If the child keyring's OnEncrypt fails, this OnEncrypt MUST also fail." I think a non-fatal exception is currently just handled by keyrings by returning the original encryption materials instead of throwing an actual exception.

src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
WesleyRosenblum: OnDecrypt catches all exceptions and only fails if every keyring failed.

src/main/java/com/amazonaws/encryptionsdk/DefaultCryptoMaterialsManager.java
```diff
@@ -13,131 +15,185 @@
...
- * This default implementation delegates to a specific {@link MasterKeyProvider} specified at construction time. It also
- * handles generating trailing signature keys when needed, placing them in the encryption context (and extracting them
- * at decrypt time).
+ * This default implementation delegates to a specific {@link MasterKeyProvider} or {@link Keyring} specified at
+ * construction time. It also handles generating trailing signature keys when needed, placing them in the
+ * encryption context (and extracting them at decrypt time).
  */
 public class DefaultCryptoMaterialsManager implements CryptoMaterialsManager {
+    private final Keyring keyring;
     private final MasterKeyProvider<?> mkp;
```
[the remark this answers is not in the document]
WesleyRosenblum: Added

src/main/java/com/amazonaws/encryptionsdk/keyrings/StandardKeyrings.java
WesleyRosenblum: I've renamed the builder methods to ...Builder

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkId.java
robin-aws: I think that meets the minimum bar. Is there a reason we can't be more specific and say "don't provide a blank string"? I'm in favour of helping out developers that have hooked things up incorrectly as specifically as possible.

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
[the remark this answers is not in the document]
robin-aws: Fair enough - as discussed offline I think there is a threshold where request objects/builders become a good idea, but something that takes a single argument is probably not worth it.

src/main/java/com/amazonaws/encryptionsdk/CreateDecryptingInputStreamRequest.java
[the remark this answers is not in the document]
robin-aws: I won't make this blocking, but for the record would object to any standards that make us repeat what is already expressed in the signature. I'll push on this independently.

src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
robin-aws: Ah you're correct, I was thinking of the `onDecrypt` behaviour here.

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkId.java
WesleyRosenblum: Validate.notBlank is checking if the string is null, whitespace only, empty string etc, which I'm not sure is widely recognized as the definition of "blank". I'm not sure I see how saying don't provide a blank string is actually more specific than saying it is required.

WesleyRosenblum: I could say "keyId must be neither null, a length of zero (no characters), empty nor whitespace"

WesleyRosenblum: Validate.notBlank is checking if the string is null, whitespace only, empty string etc, which I'm not sure is widely recognized as the definition of "blank". How about I say "keyId must be neither null, a length of zero (no characters), empty nor whitespace"

src/test/java/com/amazonaws/crypto/examples/FileStreamingExampleTest.java
```diff
@@ -0,0 +1,44 @@
...
+    @Test
+    void testEncryptAndDecrypt() throws IOException {
+        final File tempFile = File.createTempFile("FileStreamingExampleTest-TempTestData", ".tmp");
+        final File encryptedFile = new File(tempFile.getPath()  + ".encrypted");
+        final File decryptedFile = new File(tempFile.getPath() + ".decrypted");
+        tempFile.deleteOnExit();
+        encryptedFile.deleteOnExit();
+        decryptedFile.deleteOnExit();
+
+        try(BufferedWriter writer = Files.newBufferedWriter(tempFile.toPath())) {
```
robin-aws:
```suggestion
        try (BufferedWriter writer = Files.newBufferedWriter(tempFile.toPath())) {
```

src/test/java/com/amazonaws/crypto/examples/FileStreamingExampleTest.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+
+class FileStreamingExampleTest {
+
+    @Test
+    void testEncryptAndDecrypt() throws IOException {
+        final File tempFile = File.createTempFile("FileStreamingExampleTest-TempTestData", ".tmp");
+        final File encryptedFile = new File(tempFile.getPath()  + ".encrypted");
```
robin-aws:
```suggestion
        final File encryptedFile = new File(tempFile.getPath() + ".encrypted");
```

src/test/java/com/amazonaws/crypto/examples/FileStreamingExampleTest.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+
+class FileStreamingExampleTest {
+
+    @Test
+    void testEncryptAndDecrypt() throws IOException {
+        final File tempFile = File.createTempFile("FileStreamingExampleTest-TempTestData", ".tmp");
```
robin-aws: try-with-resources? Or at least a try-finally? Yes it's a temp file but it's still polite to delete it. :)

src/test/java/com/amazonaws/crypto/examples/FileStreamingExampleTest.java
```diff
@@ -0,0 +1,44 @@
...
+        final File encryptedFile = new File(tempFile.getPath()  + ".encrypted");
+        final File decryptedFile = new File(tempFile.getPath() + ".decrypted");
+        tempFile.deleteOnExit();
+        encryptedFile.deleteOnExit();
+        decryptedFile.deleteOnExit();
+
+        try(BufferedWriter writer = Files.newBufferedWriter(tempFile.toPath())) {
+            for(int i = 0; i < 1000 ; i++) {
```
robin-aws:
```suggestion
            for (int i = 0; i < 1000 ; i++) {
```

src/test/java/com/amazonaws/encryptionsdk/AwsCryptoTest.java
```diff
@@ -106,7 +131,30 @@ private void doTamperedEncryptDecrypt(final CryptoAlgorithm cryptoAlg, final int
                     masterKeyProvider,
                     cipherText
                     ).getResult();
-            Assert.fail("Expected BadCiphertextException");
+            fail("Expected BadCiphertextException");
+        } catch (final BadCiphertextException ex) {
```
robin-aws: Since you're touching this... :) assertThrows?

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkId.java
robin-aws: Yes I think that's reasonable, however you want to word it. I'm just imagining someone hitting this and getting "keyId is required" and saying "but I AM passing it dammit, and I know it can't be null..." :)

src/test/java/com/amazonaws/encryptionsdk/DecryptRequestTest.java
```diff
@@ -0,0 +1,65 @@
...
+        byte[] ciphertext = new AwsCrypto().encrypt(EncryptRequest.builder()
+                .keyring(new TestKeyring("keyId"))
+                .plaintext(new byte[]{4, 5, 6})
+                .build()).getResult();
+
+
+        assertTrue(DecryptRequest.builder()
+                .keyring(keyring)
+                .ciphertext(ciphertext).build().cryptoMaterialsManager()
+                instanceof DefaultCryptoMaterialsManager);
```
robin-aws: There's a lot happening on one line here. Consider breaking up into 2 or 3 lines for readability.

src/test/java/com/amazonaws/encryptionsdk/DefaultCryptoMaterialsManagerTest.java
```diff
@@ -308,4 +423,31 @@ public void decrypt_whenTrailingSigMissing_throwsException() throws Exception {
...
+
+            DecryptionMaterialsRequest request = DecryptionMaterialsRequest.newBuilder()
+                    .setEncryptedDataKeys(encryptMaterials.getEncryptedDataKeys())
+                    .setAlgorithm(algorithm)
+                    .setEncryptionContext(Collections.emptyMap())
+                    .build();
+
+            try {
+                new DefaultCryptoMaterialsManager(keyring1).decryptMaterials(request);
+                fail("expected exception");
```
robin-aws: assertThrows for new code at least please :)

src/test/java/com/amazonaws/encryptionsdk/DefaultCryptoMaterialsManagerTest.java
```diff
@@ -275,15 +343,62 @@ public void decrypt_testSimpleRoundTrip() throws Exception {
...
+
     @Test(expected = CannotUnwrapDataKeyException.class)
-    public void decrypt_onDecryptFailure() throws Exception {
+    public void decrypt_onDecryptFailure() {
         new DefaultCryptoMaterialsManager(mock(MasterKeyProvider.class)).decryptMaterials(
                 decryptReqFromMaterials(easyGenMaterials(ignored -> {}))
         );
     }
 
+    @Test(expected = CannotUnwrapDataKeyException.class)
```
robin-aws: I suggest you cut a follow-up PR to use assertThrows everywhere, actually.

src/test/java/com/amazonaws/encryptionsdk/kms/AwsKmsCmkIdTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @Test
+    void testToString() {
+        assertEquals(VALID_ARN, AwsKmsCmkId.fromString(VALID_ARN).toString());
+    }
+
+    @Test
+    void testEquals() {
+        assertEquals(AwsKmsCmkId.fromString(VALID_ARN), AwsKmsCmkId.fromString(VALID_ARN));
+        assertNotEquals(AwsKmsCmkId.fromString(VALID_ALIAS), AwsKmsCmkId.fromString(VALID_ALIAS_ARN));
+
```
robin-aws:
```suggestion
```

src/test/java/com/amazonaws/crypto/examples/FileStreamingExampleTest.java
WesleyRosenblum: I'm using deleteOnExit() a couple lines down, is that not enough?

src/test/java/com/amazonaws/encryptionsdk/DefaultCryptoMaterialsManagerTest.java
WesleyRosenblum: I'd been avoiding using JUnit5 assertions in JUnit 4 tests, but now that you mention it I don't there is a reason to do so

src/test/java/com/amazonaws/encryptionsdk/DefaultCryptoMaterialsManagerTest.java
WesleyRosenblum: #157

src/test/java/com/amazonaws/encryptionsdk/DefaultCryptoMaterialsManagerTest.java
robin-aws: I'm happy if you want to deal with this in #157 if you're wary

src/test/java/com/amazonaws/crypto/examples/FileStreamingExampleTest.java
robin-aws: Oh I somehow managed to miss that. :)

I have some bad experience with `deleteOnExit` in long-running processes (leaking memory by building up a never-ending list of files to delete) so I wouldn't recommend using it in any library source code, but I think it's fine for tests.

PR #152, opened 2020-01-18
Using original Materials classes instead of new Keyring classes

WesleyRosenblum:
*Issue #:* #102

*Description of changes:*

Since the CryptoMaterialsManager interface uses the original
EncryptionMaterials and DecryptionMaterials, we would force customers
to update their code once we deprecate those materials in favor of the
new Keyring versions. To avoid this, Keyrings will now use modified
versions of the original Materials classes, with certain methods
deprecated.

In addition, this change removes the MasterKeyProviderKeyring, which
was found to not be feasible for the Java ESDK since MasterKeys are
provided in the result object, which would force MasterKeys to be
incorporated into Keyrings.

# Check any applicable:


src/main/java/com/amazonaws/encryptionsdk/keyrings/KmsKeyring.java
```diff
@@ -77,15 +80,15 @@ public void onEncrypt(EncryptionMaterials encryptionMaterials) {
...
-        if (!encryptionMaterials.hasPlaintextDataKey() && generatorKeyId == null) {
+        if (encryptionMaterials.getCleartextDataKey() == null && generatorKeyId == null) {
             throw new AwsCryptoException("Encryption materials must contain either a plaintext data key or a generator");
         }
 
         final List<String> keyIdsToEncrypt = new ArrayList<>(keyIds);
 
         // If the input encryption materials do not contain a plaintext data key and a generator is defined onEncrypt
         // MUST attempt to generate a new plaintext data key and encrypt that data key by calling KMS GenerateDataKey.
-        if (!encryptionMaterials.hasPlaintextDataKey()) {
```
SalusaSecondus: While not a major issue, I like having `hasPlaintextDataKey` as it means we can check for presence without dropping more copies of a cryptographic key throughout our memory.

src/main/java/com/amazonaws/encryptionsdk/keyrings/MultiKeyring.java
```diff
@@ -50,9 +52,9 @@ public void onEncrypt(EncryptionMaterials encryptionMaterials) {
             generatorKeyring.onEncrypt(encryptionMaterials);
         }
 
-        if (!encryptionMaterials.hasPlaintextDataKey()) {
```
SalusaSecondus: Ditto on my liking `hasPlaintextDataKey()`

src/main/java/com/amazonaws/encryptionsdk/keyrings/KmsKeyring.java
WesleyRosenblum: Yeah I like it too, it was more necessary in the other materials classes since I was throwing an InvalidStateException if the plaintext data key was null. I removed it since it wasn't strictly necessary anymore, but it does improve readability somewhat. I'll add it back

PR #154, opened 2020-02-05
Making tests opt-out instead of opt-in and update TestVectorRunner

WesleyRosenblum:
Description of changes:

JUnit5 doesn't support test suites yet (see https://github.com/junit-team/junit5/issues/744) and the existing
test suites do not support the new JUnit5 tests that are being used for keyrings. This change removes the test
suites, and configures Maven to include all tests except those marked with certain JUnit tags.

Additionally, this change updates the TestVectorRunner to also test Keyrings and removes the redundant XCompat tests.

# Check any applicable:


pom.xml
```diff
@@ -208,30 +208,50 @@
                         <artifactId>maven-surefire-plugin</artifactId>
                         <version>2.22.0</version>
                         <configuration>
-                            <includes>
-                                <include>**/AllTestsSuite.java</include>
-                            </includes>
+                            <excludedGroups>slow</excludedGroups>
```
mattsb42-aws: "full" would seem to indicate that we're running everything. If we're not running slow tests in "full", where are we running them?

WesleyRosenblum: The only test marked as slow is FrameEncryptionHandlerVeryLongTest, which takes 3 hours to run. I'm not sure under what circumstances we would want to run it, but you can always manually run it. Would it be better if I renamed this tag to something like manualInvocationOnly?

SalusaSecondus: Are we stuck with the name "full-test-suite"? Why not just "test-suite"?

src/test/java/com/amazonaws/encryptionsdk/TestVectorRunner.java
```diff
@@ -126,21 +139,15 @@ private static InputStream readFromJar(JarFile jar, String name) throws IOExcept
...
     @SuppressWarnings("unchecked")
     private static TestCase parseTest(String testName, Map<String, Object> data, Map<String, KeyEntry> keys,
-                                      JarFile jar, KmsMasterKeyProvider kmsProv) throws IOException {
+                                      JarFile jar) {
         final String plaintextUrl = (String) data.get("plaintext");
-        cacheData(jar, plaintextUrl);
         final String ciphertextURL = (String) data.get("ciphertext");
-        cacheData(jar, ciphertextURL);
+        cachedData.putIfAbsent(plaintextUrl, readBytesFromJar(jar, plaintextUrl));
+        cachedData.putIfAbsent(ciphertextURL, readBytesFromJar(jar, ciphertextURL));
```
SalusaSecondus: This always reads the data from the JAR rather than only reading it if there is a cache miss. That might be expensive as I'm not sure how the data is being cached locally by Java. You could use `computeIfAbsent` if you want though.

src/test/java/com/amazonaws/encryptionsdk/kms/KMSProviderBuilderIntegrationTests.java
```diff
@@ -115,25 +118,27 @@ public void whenConstructedWithoutArguments_canUseMultipleRegions() throws Excep
...
-                    new AwsCrypto().encryptData(
-                            KmsMasterKeyProvider.builder()
-                                                .withKeysForEncryption(key)
-                                                .build(),
-                            new byte[1]
-                    ).getResult();
-
-            new AwsCrypto().decryptData(mkp, ciphertext);
-        }
+        assertThrows(CannotUnwrapDataKeyException.class, () -> {
```
SalusaSecondus: This looks to be a bug in the original test, however I believe that the for loop over the keys should be inside the `assertThrows`.

WesleyRosenblum: Its testing that when decryptData is called on a ciphertext containing a key in the non-default region it fails. The KMSTestFixtures contains one key in the default region and one in the non-default region, so only the second one fails. So the assertThrows needs to be outside the loop since only one key ID is expected to fail

src/test/java/com/amazonaws/encryptionsdk/TestVectorRunner.java
WesleyRosenblum: Fixed

pom.xml
WesleyRosenblum: I've renamed it to test-suite. I've also renamed the tag to be "ad_hoc", as I felt it was a little confusing having a "slow" tag and also a fastTestsOnly system property.

Issue #160, opened 2020-02-26
RSA Keyring Padding modes


seebees:
Keyring padding should be an enum
for a number of reasons.

Keyrings should *only* support https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/raw-rsa-keyring.md#supported-padding-schemes

The current keyring implelemtation reuses the JCE Master key:
https://github.com/aws/aws-encryption-sdk-java/blob/keyring/src/main/java/com/amazonaws/encryptionsdk/keyrings/RawRsaKeyring.java#L32
https://github.com/aws/aws-encryption-sdk-java/blob/keyring/src/main/java/com/amazonaws/encryptionsdk/internal/RsaJceKeyCipher.java#L41

Which is overly permissive.
Also, having a nice type ahead developer experience in the IDE 
as well as being able to link to a simple enum from the documentation.

PR #162, opened 2020-03-09
Define an enum for the RSA padding scheme

WesleyRosenblum:
Description of changes:

Define an enum for the RSA padding scheme #160 

# Check any applicable:


src/main/java/com/amazonaws/encryptionsdk/keyrings/RawRsaKeyringBuilder.java
```diff
@@ -97,6 +97,31 @@ public RawRsaKeyringBuilder wrappingAlgorithm(String wrappingAlgorithm) {
      * @return The {@link Keyring} instance
      */
     public Keyring build() {
-        return new RawRsaKeyring(keyNamespace, keyName, publicKey, privateKey, wrappingAlgorithm);
+        return new RawRsaKeyring(keyNamespace, keyName, publicKey, privateKey, paddingScheme);
+    }
+
+    public enum RsaPaddingScheme {
```
SalusaSecondus: NitPick: For ease of use, calculate the correct OAEPParameterSpec instances for each of these (they are immutable) and just pull them from the enum.

src/main/java/com/amazonaws/encryptionsdk/keyrings/RawRsaKeyringBuilder.java
```diff
@@ -97,6 +97,31 @@ public RawRsaKeyringBuilder wrappingAlgorithm(String wrappingAlgorithm) {
...
+
+        private final String transformation;
+
+        RsaPaddingScheme(String transformation) {
+            this.transformation = transformation;
+        }
+
+        /**
+         * The Cipher transformation standard name as specified in
+         * https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#Cipher
```
SalusaSecondus: This is *almost* correct. Please clarify that in all cases the hash function used with MGF1 is the same as the hash function used directly with the message.

WesleyRosenblum: Updated

src/main/java/com/amazonaws/encryptionsdk/keyrings/RawRsaKeyringBuilder.java
WesleyRosenblum: I'll hold off on this until Master Keys are deleted

PR #163, opened 2020-03-11
Making client suppliers composable

WesleyRosenblum:
Description of changes:

Making the Aws KMS client suppliers composable to match the design of the Python and JS ESDKs

# Check any applicable:


src/examples/java/com/amazonaws/crypto/examples/datakeycaching/MultiRegionRecordPusherExample.java
```diff
@@ -73,10 +72,10 @@ public MultiRegionRecordPusherExample(final Region[] regions, final String kmsAl
                     .build());
 
             keyrings.add(StandardKeyrings.awsKmsBuilder()
-                    .awsKmsClientSupplier(AwsKmsClientSupplier.builder()
+                    .awsKmsClientSupplier(new AllowRegionsAwsKmsClientSupplier(Collections.singleton(region.getName()),
```
seebees: You have all of these nice builders, why is this one `new`?

WesleyRosenblum: Seemed like it might be overkill to have a builder here, but I can make one too. I wasn't really satisfied with the mix of builders and constructors either

src/main/java/com/amazonaws/encryptionsdk/kms/StandardAwsKmsClientSuppliers.java
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Builder to construct an AwsKmsClientSupplier that will create and cache clients
+     * for any region. CredentialProvider and ClientConfiguration are optional and may
+     * be configured if necessary.
+     */
+    public static class DefaultAwsKmsClientSupplierBuilder {
+
+        private AWSCredentialsProvider credentialsProvider;
+        private ClientConfiguration clientConfiguration;
+        private final Map<String, AWSKMS> clientsCache = new HashMap<>();
```
SalusaSecondus: This doesn't look to be thread-safe. Should this be a `ConcurrentHashMap`?

src/main/java/com/amazonaws/encryptionsdk/kms/StandardAwsKmsClientSuppliers.java
```diff
@@ -0,0 +1,252 @@
...
+         * @param client   The client to proxy
+         * @param regionId The region the client is associated with
+         * @return The proxy
+         */
+        private AWSKMS newCachingProxy(AWSKMS client, String regionId) {
+            return (AWSKMS) Proxy.newProxyInstance(
+                    AWSKMS.class.getClassLoader(),
+                    new Class[]{AWSKMS.class},
+                    (proxy, method, methodArgs) -> {
+                        try {
```
SalusaSecondus: Is it worth remembering if we've successfully put it in the cache, and if so, short-circuit all of this logic and just delegate the call without the additional logic?

WesleyRosenblum: I'm not sure I understand, the code already short-circuits on line 99 if the client is in the cache

SalusaSecondus: Yes. But for the first client supplied, a proxy is returned with this more expensive logic which is used every time.

Of course, if the KMS Key Ring always requests a new client for _every call_, then this is a non-issue.

WesleyRosenblum: Got it. Yes the KMS Key ring requests a new client every call, it doesn't maintain any reference to the client it gets.

src/main/java/com/amazonaws/encryptionsdk/kms/StandardAwsKmsClientSuppliers.java
WesleyRosenblum: Good point, I'll make it ConcurrentHashMap. The RegionalClientSupplier that MKPs use also used a ConcurrentHashMap

PR #166, opened 2020-03-27
fix: The final frame can not be larger than the Frame Length

WesleyRosenblum:
Add validation to ensure the length of the final frame in the final frame header does not exceed the frame size specified in the message header.

See https://github.com/aws/aws-encryption-sdk-javascript/pull/281

# Check any applicable:

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
@@ -119,6 +119,11 @@ class FrameDecryptionHandler implements CryptoHandler {
             if (currentFrameHeaders_ == null) {
                 currentFrameHeaders_ = new CipherFrameHeaders();
                 currentFrameHeaders_.setNonceLength(nonceLen_);
+                if (frameSize_ == 0) {
+                    // if frame size in ciphertext headers is 0, the frame size
+                    // will need to be parsed in individual frame headers.
+                    currentFrameHeaders_.includeFrameSize(true);
+                }
             }
 
             totalParsedBytes += currentFrameHeaders_.deserialize(bytesToParse, totalParsedBytes);
@@ -130,7 +135,7 @@ class FrameDecryptionHandler implements CryptoHandler {
                     protectedContentLen = currentFrameHeaders_.getFrameContentLength();
 
                     // The final frame should not be able to exceed the frameLength
-                    if(protectedContentLen > frameSize_) {
+                    if(frameSize_ > 0 && protectedContentLen > frameSize_) {
                         throw new BadCiphertextException("Final frame length exceeds frame length.");
                     }
                 } else {
--- a/src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
@@ -135,7 +135,7 @@ class FrameDecryptionHandler implements CryptoHandler {
                     protectedContentLen = currentFrameHeaders_.getFrameContentLength();
 
                     // The final frame should not be able to exceed the frameLength
-                    if(frameSize_ > 0 && protectedContentLen > frameSize_) {
+                    if (frameSize_ > 0 && protectedContentLen > frameSize_) {
                         throw new BadCiphertextException("Final frame length exceeds frame length.");
                     }
                 } else {
--- a/src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
@@ -133,6 +133,11 @@ class FrameDecryptionHandler implements CryptoHandler {
                 int protectedContentLen = -1;
                 if (currentFrameHeaders_.isFinalFrame()) {
                     protectedContentLen = currentFrameHeaders_.getFrameContentLength();
+
+                    // The final frame should not be able to exceed the frameLength
+                    if(frameSize_ > 0 && protectedContentLen > frameSize_) {
+                        throw new BadCiphertextException("Final frame length exceeds frame length.");
+                    }
                 } else {
                     protectedContentLen = frameSize_;
                 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
```diff
@@ -133,6 +133,11 @@ public ProcessingSummary processBytes(final byte[] in, final int off, final int
                 int protectedContentLen = -1;
                 if (currentFrameHeaders_.isFinalFrame()) {
                     protectedContentLen = currentFrameHeaders_.getFrameContentLength();
+
+                    // The final frame should not be able to exceed the frameLength
+                    if(frameSize_ > 0 && protectedContentLen > frameSize_) {
```
seebees: frame size can be 0 and the final frame content length can be strictly equal to the frame size.

WesleyRosenblum: I'm only throwing the exception if the final frame size is strictly greater than the frame size. If the frame size is 0, then I can't compare the last frame size so I don't throw the exception in that case.

seebees: According to https://github.com/aws/aws-encryption-sdk-java/blob/163fd8847760a0f3e3af473530f8f7ccf03ae910/src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java#L122 frameSize_ _can_ be 0, but from https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/message-format.html#header-frame-length it is not the intention of the frameLength to be 0 in the header.

WesleyRosenblum: @SalusaSecondus do you know if there a reason we allow the frame size in the header to be 0?

mattsb42-aws: This code *should* never execute if the frame length is 0.

encryption:
https://github.com/aws/aws-encryption-sdk-java/blob/163fd8847760a0f3e3af473530f8f7ccf03ae910/src/main/java/com/amazonaws/encryptionsdk/internal/EncryptionHandler.java#L120-L146

decryption:
https://github.com/aws/aws-encryption-sdk-java/blob/163fd8847760a0f3e3af473530f8f7ccf03ae910/src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java#L497-L505

The comment on the quoted lines makes me think that this was something that was added in a long time ago, possibly for a very early version of the message format because the frame size is not included in the frame header.

https://github.com/aws/aws-encryption-sdk-java/blob/163fd8847760a0f3e3af473530f8f7ccf03ae910/src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java#L123-L124

WesleyRosenblum: Is it possible for the ciphertext to be corrupted/manipulated in some way that leaves the frame size as 0, or would that be caught upstream and cause decryption to fail?

WesleyRosenblum:
> We also need a check on the encrypt path. Arguably, that's the more important path to cover.

I thought the decrypt path was needed because we are deserializing data that may have been manipulated or corrupted somehow. The encrypt path is in our control so as long we trust ourselves to write the correct size frame for the final frame I don't see why the check is necessary?

src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
WesleyRosenblum: I've added validation to ensure the frame size is not zero if the data is framed

src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
```diff
@@ -119,11 +119,6 @@ public ProcessingSummary processBytes(final byte[] in, final int off, final int
             if (currentFrameHeaders_ == null) {
                 currentFrameHeaders_ = new CipherFrameHeaders();
                 currentFrameHeaders_.setNonceLength(nonceLen_);
-                if (frameSize_ == 0) {
-                    // if frame size in ciphertext headers is 0, the frame size
-                    // will need to be parsed in individual frame headers.
-                    currentFrameHeaders_.includeFrameSize(true);
-                }
```
SalusaSecondus, line 126: Why are we removing this logic?

src/main/java/com/amazonaws/encryptionsdk/model/CiphertextHeaders.java
```diff
@@ -143,6 +143,10 @@ public CiphertextHeaders(final byte version, final CiphertextType type, final Cr
         messageId_ = new byte[Constants.MESSAGE_ID_LEN];
         RND.nextBytes(messageId_);
 
+        if(contentType == ContentType.FRAME && frameSize <= 0) {
```
SalusaSecondus, line 146: Space after `if`

src/main/java/com/amazonaws/encryptionsdk/model/CiphertextHeaders.java
```diff
@@ -630,6 +634,10 @@ public int deserialize(final byte[] b, final int off) throws ParseException {
                 parsedBytes += parseHeaderTag(b, off + parsedBytes);
             }
 
+            if(frameLength_ <= 0 && ContentType.deserialize(contentTypeVal_) == ContentType.FRAME) {
```
SalusaSecondus, line 637: Space after `if`

src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
WesleyRosenblum: After discussion, I'll revert this part of the change and have created a separate issue to track removing this code (which was a partially implemented variable frame length feature): https://github.com/aws/aws-encryption-sdk-java/issues/167

src/main/java/com/amazonaws/encryptionsdk/internal/FrameDecryptionHandler.java
WesleyRosenblum: From discussing with @SalusaSecondus, frame size = 0 in the header was intended to indicate a variable length encoding. This feature was not fully implemented and is not documented, so it should be removed, but there are more remnants of it scattered in other parts of the code, so I've opened #167 to track that. For this PR I will only validate the final frame length if the header frame length is > 0

PR #178, opened 2020-04-17
Add an example for replicating the behavior of the AWS KMS MKP with the AWS KMS keyring

WesleyRosenblum:
*Issue #, if available:*

*Description of changes:*

In addition to using the configured AWS KMS CMKs, the AWS KMS master key provider attempts to decrypt any data keys that were encrypted under an AWS KMS CMK. This behavior is not always obvious, and when we designed the AWS KMS keyring, we decided that it was best to separate these behaviors, but as customers migrate from master key providers to keyrings, it will be helpful to have an example of how to replicate the exact behavior of the AWS KMS master key provider using keyrings.

# Check any applicable:

### DIFF
--- a/src/examples/java/com/amazonaws/crypto/examples/keyring/awskms/ActLikeAwsKmsMasterKeyProvider.java
+++ b/src/examples/java/com/amazonaws/crypto/examples/keyring/awskms/ActLikeAwsKmsMasterKeyProvider.java
@@ -1,0 +1,114 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.amazonaws.crypto.examples.keyring.awskms;
+
+import com.amazonaws.encryptionsdk.AwsCrypto;
+import com.amazonaws.encryptionsdk.AwsCryptoResult;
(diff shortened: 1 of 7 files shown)


src/examples/java/com/amazonaws/crypto/examples/keyring/awskms/ActLikeAwsKmsMasterKeyProvider.java
```diff
@@ -0,0 +1,143 @@
...
+        final Keyring cmkKeyring = StandardKeyrings.awsKmsBuilder()
+                .generatorKeyId(awsKmsCmk)
+                .keyIds(awsKmsAdditionalCmks)
+                .build();
+
+        // Create an AWS KMS discovery keyring that will attempt to decrypt
+        // any data keys that were encrypted under an AWS KMS CMK.
+        final Keyring discoveryKeyring = StandardKeyrings.awsKmsDiscoveryBuilder().build();
+
+        // Combine the single-CMK and discovery keyrings
```
mattsb42-aws, line 97: whoops; I missed this in the Python example too; I'll send out a PR to fix that

```suggestion
        // Combine the CMK and discovery keyrings
```

PR #191, opened 2020-05-26
fix: validate entire ciphertext has been processed before returning

WesleyRosenblum:
*Issue #, if available:*

*Description of changes:*

Validate that the entire ciphertext has been processed before returning

# Check any applicable:

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/internal/BlockDecryptionHandler.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/internal/BlockDecryptionHandler.java
@@ -13,15 +13,14 @@
 
 package com.amazonaws.encryptionsdk.internal;
 
-import java.util.Arrays;
-
-import javax.crypto.Cipher;
-import javax.crypto.SecretKey;
-
 import com.amazonaws.encryptionsdk.CryptoAlgorithm;
 import com.amazonaws.encryptionsdk.exception.AwsCryptoException;
 import com.amazonaws.encryptionsdk.exception.BadCiphertextException;
 import com.amazonaws.encryptionsdk.model.CipherBlockHeaders;
+
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import java.util.Arrays;
 
 /**
  * The block decryption handler is an implementation of CryptoHandler that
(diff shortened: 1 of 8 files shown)


src/main/java/com/amazonaws/encryptionsdk/internal/BlockDecryptionHandler.java
```diff
@@ -13,16 +13,15 @@
 
 package com.amazonaws.encryptionsdk.internal;
 
-import java.util.Arrays;
```
seebees, line 16: nit: Why the noop diff?

WesleyRosenblum: My IDE reordered them alphabetically

src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java
```diff
@@ -326,6 +326,10 @@ public int doFinal(final byte[] out, final int outOff) throws BadCiphertextExcep
         } else {
             int result = contentCryptoHandler_.doFinal(out, outOff);
 
+            if (!ciphertextHeaders_.isComplete() || !contentCryptoHandler_.isComplete()) {
+                throw new BadCiphertextException("Unable to process entire ciphertext.");
+            }
```
SalusaSecondus, line 331: Shouldn't the `return 0` on line 325 throw this exception also?

WesleyRosenblum: There are a couple streaming tests decrypting empty files that would fail if I change that:
https://github.com/aws/aws-encryption-sdk-java/blob/f30d81288252d446515ef4b25d6ac70b7ec64b49/src/test/java/com/amazonaws/encryptionsdk/CryptoInputStreamTest.java#L448-L459
https://github.com/aws/aws-encryption-sdk-java/blob/f30d81288252d446515ef4b25d6ac70b7ec64b49/src/test/java/com/amazonaws/encryptionsdk/CryptoOutputStreamTest.java#L451-L464

We specifically check for zero length input here: https://github.com/aws/aws-encryption-sdk-java/blob/2246cbf856a7713c0e2e37105bf90405305a628d/src/main/java/com/amazonaws/encryptionsdk/internal/DecryptionHandler.java#L226-L228

Do we not want to continue supporting that?

SalusaSecondus: Streaming returning empty for a completely empty ciphertext is valid.
Streaming should throw if any amount of invalid ciphertext is present.
One-shot should fail for anything except a complete valid ciphertext.

(Your last link) It is okay to process zero bytes while handling a larger amount of data. However we need to ensure the proper behavior when exposed to our callers.

SalusaSecondus: Sorry. You are completely correct and I see the tests to verify correct behavior. This all looks good to me.

PR #231, opened 2020-12-17
chore: Add AWS CodeBuild buildspec

seebees:
Adding buildspec to batch build
multiple JDK environments

* corretto8
* corretto11
* openjdk8
* openjdk11

# Check any applicable:

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -1,0 +1,25 @@
+version: 0.2
+
+batch:
+  fast-fail: false
+  build-list:
+    - identifier: openjdk8
+      buildspec: codebuild/openjdk8.yml
+      env:
+        env:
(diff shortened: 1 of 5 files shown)


buildspec.yml
```diff
@@ -0,0 +1,25 @@
+version: 0.2
+
+batch:
+  fast-fail: false
+  build-list:
+    - identifier: openjdk8
```
farleyb-amazon, line 6: For discussion: with Travis we were also checking openjdk9 and openjdk10. Do we want to bring them in here as well?

seebees: I think that this is a good callout.
I would like to get this merged as it is additive right now.

farleyb-amazon (approved):

PR #235, opened 2021-01-15
chore: Clarify KmsMasterKeyProvider Strict behavior in docs

lavaleri:
*Issue #, if available:*

*Description of changes:* We've seen confusion in other SDKs about this behavior. In Java we state pretty clearly this in the examples, but do not mention it directly in the KmsMasterKeyProvider docs. This fixes that.

# Check any applicable:


lavaleri:
Note from June that saying something like "configured ARN" could be confusing for customers, as they could then think that a "Configured ARN" is a thing they need to understand. Instead, consider rewording to make it explicitly "the ARN you configured the MKP with".

farleyb-amazon (approved):

PR #241, opened 2021-04-30
chore: Upgrade BouncyCastle version

robin-aws:

# Check any applicable:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcprov-ext-jdk15on</artifactId>
-            <version>1.68</version>
+            <version>1.67</version>
         </dependency>
 
         <dependency>


pom.xml
```diff
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcprov-ext-jdk15on</artifactId>
-            <version>1.65</version>
+            <version>([1.67,)</version>
```
lizroth, line 52: Per our out of band discussion, I think it's right for us to change to a fixed BC version now, and do a more thoughtful evaluation for changing to a range version instead

karlw00t: I agree with Liz. Can we scope this down to a version that we know works? Then we can make sure that doing an open ended version is a safe in a separate tasks. Do we feel confident that we have unit tests the exercise the parts of Bouncy Castle we do use, so if those behaviors change, we'll see it. Are we going to run each version of BC in our CI before we release?

robin-aws: See the latest commit, we already agreed out of band that we weren't comfortable switching to a range just yet, and we're sticking to the minimum patched version (1.67).

karlw00t (approved):

PR #244, opened 2021-05-10
chore: Automate staging and production release using CodeBuild as well

robin-aws:
*Description of changes:*

This fills in the no-op steps to publish artifacts to either staging (i.e. CodeArtifact) or production (i.e. Sonatype).

*Testing:*

The staging workflow works completely with an existing release (if not already present on CodeArtifact), and the production workflow fails as expected because the latest version is already publicly released.

# Check any applicable:

### DIFF
--- a/codebuild/release/release-prod.yml
+++ b/codebuild/release/release-prod.yml
@@ -5,7 +5,7 @@ env:
     GPG_KEY: Maven-GPG-Keys-Credentials:Keyname
     GPG_PASS: Maven-GPG-Keys-Credentials:Passphrase
     SONA_USERNAME: Sonatype-Team-Account:Username 
-    SONA_PASS: Sonatype-Team-Account:Password
+    SONA_PASSWORD: Sonatype-Team-Account:Password
 
 phases:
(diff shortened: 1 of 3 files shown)


farleyb-amazon (approved):

codebuild/release/release-prod.yml
```diff
@@ -2,13 +2,15 @@ version: 0.2
...
+    GPG_PASS: Maven-GPG-Keys-Credentials:Passphrase
     SONA_USERNAME: Sonatype-Team-Account:Username 
-    SONA_PASS: Sonatype-Team-Account:Password
+    SONA_PASSWORD: Sonatype-Team-Account:Password
 
 phases:
   install:
     runtime-versions:
-      java: openjdk8
+      java: openjdk11
```
farleyb-amazon, line 13: For posterity, per offline discussion: since the validation in these release steps is testing against all supported versions of Java (8 and 11, corretto and openjdk), bumping to 11 for the actual building of artifacts is a reasonable update.

PR #345, opened 2021-09-22
chore: source controlling cfn templates that will be used for our release process

josecorella:
*Description of changes:*
code_artifact.yml and code_build_parameter_map.json are the same files that live in the [cfn](https://github.com/aws/private-aws-encryption-sdk-java-staging/tree/infra/cfn) branch of the private repos 

# Check any applicable:

### DIFF
--- a/cfn/prod-release.yml
+++ b/cfn/prod-release.yml
@@ -1,0 +1,264 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  Template to build a CodeBuild Project, assumes that GitHub credentials are
+  already set up.
+Parameters:
+  ProjectName:
+    Type: String
+    Description: The name of the CodeBuild Project
+    Default: java-esdk-prod
+  ProjectDescription:
+    Type: String
+    Description: The description for the CodeBuild Project
+    Default: CFN stack for managing CodeBuild Release project for the ESDK-Java
+  SourceLocation:
+    Type: String
+    Description: The https GitHub URL for the project
+    Default: "https://github.com/aws/aws-encryption-sdk-java.git"
+  NumberOfBuildsInBatch:
+    Type: Number
+    MaxValue: 100
+    MinValue: 1
+    Default: 10
+    Description: The number of builds you expect to run in a batch
+Metadata:
+  "AWS::CloudFormation::Interface":
+    ParameterGroups:
+      - Label:
+          default: Crypto Tools CodeBuild Project Template
+        Parameters:
+          - ProjectName
+          - ProjectDescription
+          - SourceLocation
+Resources:
+  CodeBuildProjectRelease:
+    Type: "AWS::CodeBuild::Project"
+    Properties:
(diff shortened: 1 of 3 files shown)


cfn/prod-release.yml
```diff
@@ -0,0 +1,264 @@
...
+        - !Ref CryptoToolsKMS
+        - !Ref CodeBuildBatchPolicy
+        - !Ref CodeBuildBasePolicy
+        - !Ref SecretsManagerPolicy
+        - !Ref ParameterStorePolicy
+        - "arn:aws:iam::aws:policy/AWSCodeArtifactReadOnlyAccess"
+        - "arn:aws:iam::aws:policy/AWSCodeArtifactAdminAccess"
+    Metadata:
+      "AWS::CloudFormation::Designer":
+        id: 19b17bdc-5d49-4f12-93e6-6d761b7ce324
```
texastony, line 99: nit: Drop this

cfn/prod-release.yml
```diff
@@ -0,0 +1,264 @@
...
+                "codebuild:StartBuild",
+                "codebuild:StopBuild",
+                "codebuild:RetryBuild"
+              ]
+            }
+          ]
+        }
+    Metadata:
+      "AWS::CloudFormation::Designer":
+        id: 894a45c5-5dc7-4f11-be65-8bf5e37e289d
```
texastony, line 127: Nit: drop this

cfn/prod-release.yml
```diff
@@ -0,0 +1,264 @@
...
+              ],
+              "Resource": [
+                "arn:aws:codebuild:${AWS::Region}:${AWS::AccountId}:report-group/${ProjectName}-*"
+              ]
+            }
+          ]
+        }
+    Metadata:
+      "AWS::CloudFormation::Designer":
+        id: 3dafd088-2792-4a41-b612-d9d049721644
```
texastony, line 183: nit: just do a text search for "AWS::CloudFormation::Designer" and drop these.

They are tags for the icons in the designer tool. They do not encode any information that is relevant to the CFN stack, and they will probably change if this is ever loaded into the designer again.

texastony (changes requested):
Two thoughts:
1. There are no changes to the current infrastructure, correct? This is just documenting it.
2. Remove the metadata fields for cloud designer. 

Cloud Designer is a great tool for validating that the template is good. But it does add these meta tags that are not actually part of the stack.

josecorella:
> Two thoughts:
> 
>     1. There are no changes to the current infrastructure, correct? This is just documenting it.
> 
>     2. Remove the metadata fields for cloud designer.
> 
> 
> Cloud Designer is a great tool for validating that the template is good. But it does add these meta tags that are not actually part of the stack.

1. No changes to the infrastructure. We had talked about adding the webhook settings here in the cfn but that will come in the next iteration of updates to our ci/cd process. Once we are comfortable with pressing the button to run the release we can see about adding the webhook.

texastony (approved):
Very nice!

PR #354, opened 2021-09-28
chore: updating prod template to remove resources and abstract ones

josecorella:

# Check any applicable:

### DIFF
--- a/cfn/prod-release.yml
+++ b/cfn/prod-release.yml
@@ -107,8 +107,7 @@ Resources:
             {
               "Effect": "Allow",
               "Resource": [
-                "arn:aws:codebuild:${AWS::Region}:${AWS::AccountId}:project/${ProjectName}-test-release",
-                "arn:aws:codebuild:${AWS::Region}:${AWS::AccountId}:project/${ProjectName}-prod-release",
+                "arn:aws:codebuild:${AWS::Region}:${AWS::AccountId}:project/${ProjectName}-release-prod",
                 "arn:aws:codebuild:${AWS::Region}:${AWS::AccountId}:project/${ProjectName}"
               ],
               "Action": [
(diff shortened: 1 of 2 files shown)


cfn/prod-release.yml
```diff
@@ -212,8 +209,8 @@ Resources:
             {
               "Effect": "Allow",
               "Resource": [
-                "arn:aws:kms:*:658956600833:key/*",
-                "arn:aws:kms:*:658956600833:alias/*"
+                "arn:aws:kms:*:${AWS::AccountId}:key/*",
```
texastony, line 212: Nope! `658956600833` has the keys you need to pass the tests. 
This statement grants the CodeBuild resource permission to call the key.

cfn/prod-release.yml
```diff
@@ -133,10 +132,8 @@ Resources:
               "Resource": [
                 "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/codebuild/${ProjectName}",
                 "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/codebuild/${ProjectName}:*",
-                "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/codebuild/${ProjectName}-test-release",
-                "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/codebuild/${ProjectName}-test-release:*",
```
texastony, line 137: Are you going to make another template for test-release?
There would be much repeated code... I would favor 1 template/stack having both projects as compared to a stack per project.
Then the projects can share policy documents. Otherwise, you would have to repeat your policy documents.

texastony (changes requested):

cfn/prod-release.yml
josecorella: We discussed this offline, but just to document it here. The plan is to have just one project for both test-release and prod-release called release-prod.

cfn/prod-release.yml
```diff
@@ -173,14 +173,14 @@ Resources:
     Type: "AWS::SSM::Parameter"
     Properties:
       Description: Parameter to store our account id so CodeBuild specs can access it
-      Name: /CodeBuild/AccountId
+      Name: /CodeBuild/Account
```
texastony, line 176: These need to match this:
https://github.com/aws/aws-encryption-sdk-java/blob/ad9a05c59a5c8f280da8bbd49de7d4f8b2232121/codebuild/release/validate-staging.yml#L9

which is `/CodeBuild/AccountId`

cfn/prod-release.yml
```diff
@@ -176,14 +173,14 @@ Resources:
...
```
texastony, line 176: These need to match this:
https://github.com/aws/aws-encryption-sdk-java/blob/ad9a05c59a5c8f280da8bbd49de7d4f8b2232121/codebuild/release/validate-staging.yml#L9

which is `/CodeBuild/AccountId`

texastony (changes requested):
Revert Line 176 and you are good.

cfn/prod-release.yml
josecorella: That's a little weird, shouldn't this have been caught in the stack creation process? or would this be a case where it would detect a drift. I'll go ahead and change it so they match. The name was causing it to not create the stack, and changing it to `Account` successfully created the stack.

texastony: Ohh... then nevermind. The name already exists, right? Because we have been using it.

texastony (dismissed):
Left a suggestion, but otherwise it is brilliant!

texastony (approved):

PR #412, opened 2021-11-11
chore: update gh token

josecorella:
*Description of changes:*
Uses the correct GH PAT in order to log in to github using the gh cli. gh cli is needed in order to create the release and upload the artifacts.

# Check any applicable:

### DIFF
--- a/codebuild/release/upload_artifacts.yml
+++ b/codebuild/release/upload_artifacts.yml
@@ -8,13 +8,17 @@ env:
     BRANCH: "master"
   git-credential-helper: yes
   secrets-manager:
-    GH_TOKEN: Github/aws-crypto-tools-ci-bot:personal\ access\ token\ (new\ token\ format)
+    GH_TOKEN: Github/aws-crypto-tools-ci-bot:ESDK Release Token
 
 phases:
   pre_build:
     commands:
         # get new project version
       - export VERSION=$(grep version pom.xml | head -n 1 | sed -n 's/[ \t]*<version>\(.*\)<\/version>/\1/p')
+      - git config --global user.name "aws-crypto-tools-ci-bot"
+      - git config --global user.email "no-reply@noemail.local"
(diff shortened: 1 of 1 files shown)


codebuild/release/upload_artifacts.yml
```diff
@@ -8,13 +8,17 @@ env:
...
 
 phases:
   pre_build:
     commands:
         # get new project version
       - export VERSION=$(grep version pom.xml | head -n 1 | sed -n 's/[ \t]*<version>\(.*\)<\/version>/\1/p')
+      - git config --global user.name "aws-crypto-tools-ci-bot"
+      - git config --global user.email "no-reply@noemail.local"
+      - echo $GH_TOKEN > token.txt
+      - export GH_TOKEN=
```
texastony, line 21:
```suggestion
      - export -n GH_TOKEN
```
The above assumes that terminal CodeBuild provides us is bash. If it is `zsh`, my suggestion will fail.
But it is not: https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html#build-spec.shell
And both `sh` and `bash` support the `-n` flag on `export`.

texastony (changes requested):

codebuild/release/upload_artifacts.yml
josecorella: ooh i like that better it is much cleaner. i'll test this and get back to you if we still have the same result

josecorella: big oof! unfortunately the `export -n` option returns with `export: Illegal option -n`

texastony: For the record, CodeBuild does not support `-n` on `export`. We could probably use `unset`, but testing this is manual and time consuming, so I am approving this as is.

texastony (approved):

PR #413, opened 2021-11-11
chore: update mainline-1.x cd

josecorella:
*Description of changes:*
- Breaks up release prod into two builds, this way if we need to restart the artifact hunt it won't try to publish again.
- updates README
- uses the correct PAT token in upload artifacts

# Check any applicable:

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -57,7 +57,7 @@
           }],
           ["@semantic-release/git", {
             "assets": ["./CHANGELOG.md", "./pom.xml", "./README.md"],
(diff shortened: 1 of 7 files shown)


texastony (commented):
How was this change tested? Did you execute a test release?

josecorella:
> How was this change tested? Did you execute a test release?

These changes were from when we did the 1.9.1 release. I tested this in a build which I'll send you over slack.

texastony (approved):

PR #516, opened 2022-02-07
AWS SDK v2 support

smswz:
*Description of changes:*
Adding support for the new version of the AWS SDK for Java. The changes will be merged into the `aws-sdk-v2` branch to make it easier to compare changes. Once the source and tests are merged in to that branch, I will open a PR to merge `aws-sdk-v2` into `master`

# Check any applicable:
- [x] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files. _No files were moved_


josecorella (changes requested):
I am not sure if we are leaking information when we store certain variables.
For example, on this [line](https://github.com/aws/aws-encryption-sdk-java/pull/516/files?diff=split&w=0#diff-2242a0e03198f5f2578321916257fd101e193a93961de0391fffdee91469d0d6R192):

`
final String decryptResponseKeyId = decryptResponse.keyId();
`

I understand that for readability this makes more sense but I'm not sure if adding this reference leaks information?

smswz:
Hmmm, I wouldn't think a variable would cause information to be leaked. Maybe @lavaleri  or @farleyb-amazon can provide more information?

lavaleri:
The KMS Key Id is not secret. Even if it were, we would just want to be sure to minimize the time it needs to exist in memory as much as is feasible. This extra reference won't hurt since with or without this extra reference no more references to this data will exist after this method completes, and it will be ready to be GC'd by Java as appropriate.

In terms of data leakage in Java, we mainly just want to be sure that we aren't storing secret values in such that they never get GC'd, or we potentially log that information.

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKey.java
```diff
@@ -151,36 +152,40 @@ public void addGrantToken(final String grantToken) {
...
+    final GenerateDataKeyResponse gdkResponse =
         kmsClient_.generateDataKey(
-            updateUserAgent(
-                new GenerateDataKeyRequest()
-                    .withKeyId(awsKmsIdentifier_)
-                    .withNumberOfBytes(algorithm.getDataKeyLength())
-                    .withEncryptionContext(encryptionContext)
-                    .withGrantTokens(grantTokens_)));
+            GenerateDataKeyRequest.builder()
+                .overrideConfiguration(API_NAME_INTERCEPTOR)
```
josecorella, line 158: does the overrideConfig with the API_NAME_INTERCEPTOR do the work that the updateUserAgent method did?

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -44,31 +41,29 @@
...
+  private final Region defaultRegion_;
 
   public static class Builder implements Cloneable {
-    private String defaultRegion_ = getSdkDefaultRegion();
-    private Optional<KmsMasterKeyProvider.RegionalClientSupplier> regionalClientSupplier_ =
-        Optional.empty();
-    private AWSKMSClientBuilder templateBuilder_ = null;
+    private Region defaultRegion_ = getSdkDefaultRegion();
+
+    private Supplier<KmsClientBuilder> builderSupplier_ = null;
```
lavaleri, line 52: Why is this a supplier now? Is it just to simplify the logic that was previously necessary with all the cloning?

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKey.java
smswz: yep. that api disappeared in v2. According to the devs, this is the proper way to do it

pom.xml
```diff
@@ -4,7 +4,7 @@
 
     <groupId>com.amazonaws</groupId>
     <artifactId>aws-encryption-sdk-java</artifactId>
-    <version>2.3.3</version>
+    <version>2.4.0</version>
```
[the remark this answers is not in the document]
smswz: I did have to update for testing, but will revert to 2.3.3 before merging

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -44,31 +41,29 @@
...
+
     private DiscoveryFilter discoveryFilter_ = null;
-    private String discoveryMrkRegion_ = this.defaultRegion_;
+    private Region discoveryMrkRegion_ = this.defaultRegion_;
 
     Builder() {
       // Default access: Don't allow outside classes to extend this class
     }
 
     public Builder clone() {
```
lavaleri, line 62: This isn't blocking, but this is a good opportunity to move from Cloneable to a copy constructor pattern, i.e. implement a toBuilder() instead of this clone. The Cloneable pattern is confusing and easy to get wrong, and I believe the SDK already has customers familiar with the toBuilder() pattern.

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -112,19 +106,17 @@ public Builder withDiscoveryMrkRegion(
...
+     * IllegalStateException} will be thrown.
      *
-     * @see
-     *     KmsMasterKeyProvider.Builder#withCustomClientFactory(KmsMasterKeyProvider.RegionalClientSupplier)
+     * @see KmsMasterKeyProvider.Builder#customRegionalClientSupplier(RegionalClientSupplier)
      */
-    public Builder withCustomClientFactory(
-        KmsMasterKeyProvider.RegionalClientSupplier regionalClientSupplier) {
-      if (templateBuilder_ != null) {
+    public Builder customRegionalClientSupplier(RegionalClientSupplier regionalClientSupplier) {
```
lavaleri, line 114: This might be the right move regardless, but just noting that this name change is one more way that customers could potentially get confused in this migration. Any migration guide should make it clear that anyone previously specifying "customerClientFactory" should now specify "regionalClientSupplier".

src/main/java/com/amazonaws/encryptionsdk/kmsv2/KmsMasterKeyProvider.java
```diff
@@ -47,24 +39,15 @@
...
 
   public static class Builder implements Cloneable {
-    private String defaultRegion_ = null;
+    private Region defaultRegion_ = null;
+
+    private Supplier<KmsClientBuilder> builderSupplier_ = null;
     private RegionalClientSupplier regionalClientSupplier_ = null;
-    private AWSKMSClientBuilder templateBuilder_ = null;
+
+    //    private AWSKMSClientBuilder templateBuilder_ = null;
```
lavaleri, line 50: nit: remove

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -134,54 +126,19 @@ private RuntimeException clientSupplierComboException() {
...
-     * #withCredentials(AWSCredentialsProvider)}.
+     * KmsClientBuilder} to configure KMS clients. Note that the region set on this builder will be
+     * ignored, but all other settings will be propagated into the regional clients.
      *
-     * @see KmsMasterKeyProvider.Builder#withClientBuilder(AWSKMSClientBuilder)
+     * @see KmsMasterKeyProvider.Builder#builderSupplier(Supplier)
      */
-    public Builder withClientBuilder(AWSKMSClientBuilder builder) {
-      if (regionalClientSupplier_.isPresent()) {
+    public Builder builderSupplier(Supplier<KmsClientBuilder> supplier) {
```
lavaleri, line 136: I'm still not sure if this needs to be a supplier, but if it does, should we provide a convenience method that will also just accept a KmsClientBuilder to make this simpler for customers?

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -100,8 +95,7 @@ public Builder withDefaultRegion(String defaultRegion) {
      *
      * @param discoveryMrkRegion The region to contact to attempt to decrypt multi-region keys.
      */
-    public Builder withDiscoveryMrkRegion(
-        String discoveryMrkRegion) {
+    public Builder discoveryMrkRegion(Region discoveryMrkRegion) {
```
lavaleri, line 98: Do we need to support accepting a string as well? https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/region-selection.html

src/main/java/com/amazonaws/encryptionsdk/kmsv2/KmsMasterKeyProvider.java
```diff
@@ -127,83 +108,27 @@ private RuntimeException clientSupplierComboException() {
               + "withCredentials or withClientBuilder");
```
lavaleri, line 108: If we really can't support withCredentials in SDKv2, then lets do a grep and make sure we remove every reference to it.

src/main/java/com/amazonaws/encryptionsdk/kmsv2/KmsMasterKeyProvider.java
```diff
@@ -3,33 +3,25 @@
 
 package com.amazonaws.encryptionsdk.kmsv2;
```
lavaleri, line 4: We can bikeshed the naming some more, but I would prefer `kmssdkv2` to `kmsv2`, as that makes it clearer that the sdk is what is being versioned, not kms itself.

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKey.java
```diff
@@ -38,18 +40,17 @@
  */
 public final class AwsKmsMrkAwareMasterKey extends MasterKey<AwsKmsMrkAwareMasterKey>
     implements KmsMethods {
-  private static final String USER_AGENT = VersionInfo.loadUserAgent();
-  private final AWSKMS kmsClient_;
+
+  static final ApiName API_NAME =
+      ApiName.builder().name(VersionInfo.apiName()).version(VersionInfo.versionNumber()).build();
+  private static final Consumer<AwsRequestOverrideConfiguration.Builder> API_NAME_INTERCEPTOR =
+      builder -> builder.addApiName(API_NAME);
```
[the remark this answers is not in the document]
smswz: SDK v1: `AwsCrypto/2.3.3`

SDK v2: need to check during integration tests

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -134,54 +126,19 @@ private RuntimeException clientSupplierComboException() {
               + "withCredentials or withClientBuilder");
     }
 
-    /**
-     * Configures the {@link AwsKmsMrkAwareMasterKeyProvider} to use specific credentials. If a
-     * builder was previously set, this will override whatever credentials it set.
-     *
-     * @see KmsMasterKeyProvider.Builder#withCredentials(AWSCredentialsProvider)
-     */
-    public Builder withCredentials(
```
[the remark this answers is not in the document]
smswz: The customers will / should supply credentials through the KMS client builder supplier. I'll go into it a bit more with the supplier

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
smswz: The builders in SDKv2 are neither `Cloneable` nor have accessors for fields stored within them. This prevents us from copying the builders that a user provides to the library. Now, the intended path is they configure a `builderSupplier` which configures the builder in a lambda before providing it to the user. This is also how they would configure the credentials.

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
smswz: That's a great idea, especially since the SDKv2 clients no longer support the `Cloneable` pattern either. I will probably push this change to a later CR though since its more of a functional change.

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
smswz: I agree. As someone trying to figure out what was going on here, I didn't know what part of the supplier was _custom_ and I think for new users, that would hold as well. I will make sure to surface this with June.

src/main/java/com/amazonaws/encryptionsdk/kmsv2/KmsMasterKeyProvider.java
smswz: Oops! That should have been deleted!

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
smswz: It needs to be a supplier because we cannot clone the builders. If the user does not provide something here. It defaults to `KmsClient.builder()`. The can also specify this explicitly with: `.builderSupplier(KmsClient::builder)` and I intend to illustrate this through docs.

src/main/java/com/amazonaws/encryptionsdk/kmsv2/AwsKmsMrkAwareMasterKeyProvider.java
smswz: For v2, they've pretty strongly moved away from strings in place of the typed `Region`. If for some reason a user wants to use some custom region, they can use `Region.of("custom-region-1")`

src/main/java/com/amazonaws/encryptionsdk/kmsv2/KmsMasterKeyProvider.java
smswz: good call! I'll double check everything for credentials references

src/main/java/com/amazonaws/encryptionsdk/kmsv2/KmsMasterKeyProvider.java
smswz: Yea, I'd like a better name, I'm not sure 3 abbreviations in a row would be very clear though either. I'll ruminate on some ideas that are clearer.

smswz:
Dependent on https://github.com/aws/aws-sdk-java-v2/pull/3024

smswz:
After getting some eyes, I'll rename the package to `kmssdkv2`

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKey.java
```diff
@@ -90,7 +90,6 @@ private AwsKmsMrkAwareMasterKey(
           "AwsKmsMrkAwareMasterKey must be configured with an AWS KMS client.");
     }
 
-    /* Precondition: A provider is required. */
```
lavaleri, line 93: Why are we removing these? This notation and associated CI check was a precursor to our spec/duvet work, but I wouldn't want to remove these unless we could have the spec/duvet fill the gap.

src/test/java/com/amazonaws/encryptionsdk/kmsv2/XCompatKmsDecryptTest.java
```diff
@@ -3,29 +3,27 @@
...
 @RunWith(Parameterized.class)
 public class XCompatKmsDecryptTest {
-  private String plaintextFileName;
-  private String ciphertextFileName;
-  private String kmsKeyId;
+  private final String plaintextFileName;
+  private final String ciphertextFileName;
+  private final String kmsKeyId;
 
   public XCompatKmsDecryptTest(
```
lavaleri, line 28: We should add this (and MaxEncryptedDataKeysIntegrationTest.java) to https://github.com/aws/aws-encryption-sdk-java/blob/275ce526953ecf8f8f2f94ba38dcdf92e35c84bf/src/test/java/com/amazonaws/encryptionsdk/IntegrationTestSuite.java

lavaleri: That being said, I don't think we run IntegrationTestSuite anywhere in our CI, it doesn't even look like the pom defines any target for it... Historically, I think these tests were turned off before my time on the team because they just took too long. It might be worth taking another look now to determine whether we think there is actually any value for defining KMSSDKv2 versions of these tests, whether it's worth adding these tests at some point in our CI (maybe just a daily run or at least add to the release process?), or whether we should clean up the state of this repo and remove these tests if they are not in use.

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKey.java
smswz: I removed them here and added them to the v2 files because the conditions script will complain if there are duplicates.

src/test/java/com/amazonaws/encryptionsdk/kmsv2/XCompatKmsDecryptTest.java
smswz: I'll add them here, but yea, they're not specified in any target. I've run the test suites manually to make sure they pass though.

josecorella (commented):
lgtm!

josecorella (approved):

lavaleri (approved):

PR #581, opened 2022-03-10
fix(ci): Quote commit statement to prevent errant YAML parsing

smswz:
*Description of changes:*
YAML tries to parse the colon within a quoted string and causes the last step of the build to fail. Quoting the entire string prevents this condition

# Check any applicable:


josecorella (approved):
lgtm!

PR #761, opened 2022-05-11
doc: clarify JceMasterKey supported algorithms

alex-chew:

# Check any applicable:


texastony (approved):
LGTM but we would have to re-generate the Java Docs for this to be discoverable outside of the source code.

And we should NOT do that with out a release...

texastony (approved):

PR #840, opened 2022-06-09
chore: use new signing keys for ci

josecorella:
*Description of changes:*
This PR updates CI to use a different Signing Key that is only used to sign artifacts that get uploaded to CodeArtifact.

# Check any applicable:


texastony (approved):
LGTM

PR #1025, opened 2022-08-25
chore: Mark 1.x as End-of-Support

texastony:
*Issue #, if available:* Mark 1.x as End-of-Support

*Description of changes:*
- Client (AwsCrypto) emits a log warning on creation detailing 1.x is in End-of-Support
- Added the support policy from master
- updated the support policy from master to mark 1.x as End-of-Support.

# Check any applicable:

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
@@ -13,8 +13,8 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
-import java.util.logging.Logger;
 import java.util.Map;
+import java.util.logging.Logger;
 
 /**
  * Provides the primary entry-point to the AWS Encryption SDK. All encryption and decryption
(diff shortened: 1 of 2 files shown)


josecorella:
nit: this branch also gets dependabot updates, could we turn them off with this pr?

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
```diff
@@ -93,6 +95,11 @@ public class AwsCrypto {
...
     commitmentPolicy_ = CommitmentPolicy.ForbidEncryptAllowDecrypt;
     maxEncryptedDataKeys_ = CiphertextHeaders.NO_MAX_ENCRYPTED_DATA_KEYS;
+    LOGGER.warning(
+            "This major version (1.x) of the AWS Encryption SDK for Java has reached End-of-Support.\n" +
+            "It will no longer receive security updates or bug fixes.\n" +
+            "Consider updating to the latest version of the AWS Encryption SDK."
+    );
   }
 
   private AwsCrypto(Builder builder) {
```
lavaleri, line 105: It looks like this constructor and the above are separate code paths. We should also add the warning in the `AwsCrypto(Builder builder)` constructor.

If this were HEAD I'd suggest refactoring so the public constructor uses the private one, so we'd only have to add it in one place. But since this is code we are not maintaining anymore, I'm ok with just copy-pasta-ing the warning where it's needed.

lavaleri (changes requested):

lavaleri:
For dependabot, you need to update the config on the `master` branch.

josecorella:
currently codebuild is failing this pr because of:
1 - formatting errors in AwsCrypto, run `mvn com.coveo:fmt-maven-plugin:format` to resolve
2 - this branch has not been updated to use the new signing keys for ci, i'll go fix that

src/main/java/com/amazonaws/encryptionsdk/AwsCrypto.java
texastony: Fixed in 6f3e2de

josecorella:
waiting on #1027 to use the correct signing key

lavaleri (approved):

PR #1219, opened 2022-11-28
chore: Fix re-run CI workflows

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

* Modify `codebuild/ci/release-ci.yml`
  * Delete any existing CodeArtifact packages sharing the same package version (i.e. Git commit hash) before uploading package to CodeArtifact.
    * Reason: This is necessary for running/re-running workflows for the same commit hash. If previous packages are already present, the deploy step will fail. This has caused issues for recent PRs. (Every PR causes 2 runs: 1 on push, 1 on PR; the second job failed.)
    * Testing: Succeeds with both [1)](https://github.com/aws/aws-encryption-sdk-java/actions/runs/3568644437/jobs/5997748619) First run (i.e. nothing to delete); [2)](https://github.com/aws/aws-encryption-sdk-java/actions/runs/3568644437/jobs/5998204860) Subsequent runs. This is also validated by this PR's 2nd run of "Release CI" succeeding without requiring manual intervention.
    * An alternative approach I considered: skipping deployment if a package with the same commit hash is already present. I prefer the implemented approach to get a "fresh" artifact but don't feel strongly for either approach.

# Check any applicable:


codebuild/ci/release-ci.yml
```diff
@@ -27,7 +27,18 @@ phases:
...
+          --format maven \
+          --namespace $NAMESPACE \
+          --package $PACKAGE \
+          --versions $VERSION_HASH \
+          --region $REGION;
+
+      # See https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
       - echo "Setting version in POM to $VERSION_HASH"
       - mvn versions:set -DnewVersion="$VERSION_HASH" --no-transfer-progress
       - echo "Version is now $(grep version pom.xml | head -n 1 | sed -n 's/[ \t]*<version>\(.*\)<\/version>/\1/p')"
```
lavaleri, line 44: Can we update the output on L61,69 now? If Codebuild fails, it must be for some other reason now.

lucasmcdonald3: Good catch, updated

PR #1305, opened 2023-01-09
feat: Fix errors in the example code

andimarek:
There are two small errors in the example that cause compile errors:
- `.stream` instead of `.stream()`
- `data.getBytes` instead of `plaintext.getBytes`


texastony:
@ajewellamz Comment is not quite accurate.
@andimarek Our CI is currently in a bad state,
and not running for forks.
But @andimarek does not have and will not have permission to create a branch on this repo.
To unblock this merge, we will re-create this branch.

ajewellamz:
Merged as PR 1306

Issue #1768, opened 2023-09-14
AwsKmsMrkAwareMasterKeyProvider always attempts to resolve defaultRegion through DefaultAwsRegionProviderChain


spennymac:
### Problem:

After being unable to decrypt data with a multi-region key (MRK), I discovered that I needed to use the AwsKmsMrkAwareMasterKeyProvider. However, after making this change, I noticed increased latency. This was because `getDefaultRegion()` was being called every time the provider was instantiated, even when I had supplied a default region to the builder [[0]](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/kmssdkv2/AwsKmsMrkAwareMasterKeyProvider.java#L50). 

`getDefaultRegion` tries to determine the region first from the system settings, then from the AWS profile, and finally from the metadata server. I haven't configured any of these, so the increased latency was due to the call to the metadata server timing out, since I'm not operating within the AWS infrastructure.

This all leads me to a question:

Could this code change to only resolve the default region in the constructor if default regions is null? Or each `build` method of the Builder calls this if the region was not set. If yes, I'll gladly put on a PR.

Thanks in advance!

spennymac:
I understand there are many work around I can do, i just found it odd that it would attempt to resolve the the region even when you set it in the builder.

seebees:
Hi, thanks for the report.

I think I see what is going on and how to fix it.
The expense of calling `getSdkDefaultRegion()` can and should be deferred until needed.

seebees:
As of 3.0.0, the AWS Encryption SDK for Java is integrated with AWS Cryptographic Materials Provider (MPL).

The MPL features Keyrings, which give you much more granular control over when an SDK client is created.

You can replace your MasterKeyProvider instances with Keyrings, which will best resolve this issue.

PR #1964, opened 2024-01-26
fix: resolve defaultRegion only if necessary

imabhichow:
*Issue #, if available:*

- https://github.com/aws/aws-encryption-sdk-java/issues/1768

*Description of changes:*

- Defer calling `getSdkDefaultRegion()` only if no default region is provided.
- This change only resolve the default region from `DefaultAwsRegionProviderChain` if default region is null or not configured.
- Furthermore, `_defaultRegion` and `_discoveryMrkRegion` will always have non-null region values during build.

# Check any applicable:

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
@@ -194,7 +194,8 @@ public final class AwsKmsMrkAwareMasterKeyProvider
     public AwsKmsMrkAwareMasterKeyProvider buildDiscovery() {
       final boolean isDiscovery = true;
       if (discoveryMrkRegion_ == null) {
-        // The AWS SDK has a default process for evaluating the default Region. This returns null if no
+        // The AWS SDK has a default process for evaluating the default Region. This returns null if
+        // no
         // default region is found. Because a default region _may_ not be needed.
         // Note: Final Null check is at
         // com.amazonaws.encryptionsdk.kms.AwsKmsMrkAwareMasterKeyProvider#AwsKmsMrkAwareMasterKeyProvider#L417
@@ -203,9 +204,12 @@ public final class AwsKmsMrkAwareMasterKeyProvider
           // # In
           // # discovery mode if a default MRK Region is not configured the AWS SDK
           // # Default Region MUST be used.
-          discoveryMrkRegion_ = new com.amazonaws.regions.DefaultAwsRegionProviderChain().getRegion();
+          discoveryMrkRegion_ =
+              new com.amazonaws.regions.DefaultAwsRegionProviderChain().getRegion();
         } catch (SdkClientException ex) {
-          throw new AwsCryptoException("DiscoveryMrkRegion was not set and Default AWS Region Provider Chain could not load one.", ex);
+          throw new AwsCryptoException(
+              "DiscoveryMrkRegion was not set and Default AWS Region Provider Chain could not load one.",
+              ex);
         }
       }
 
@@ -421,17 +425,21 @@ public final class AwsKmsMrkAwareMasterKeyProvider
     this.regionalClientSupplier_ = supplier;
     // NPE would be thrown via result of extractRegion passed to KMS Client request around line 583.
     // We could refactor extractRegion to handle Strict and Discovery separately,
-    // which would allow us to avoid this potentially un-needed new DefaultAwsRegionProviderChain().getRegion()
+    // which would allow us to avoid this potentially un-needed new
+    // DefaultAwsRegionProviderChain().getRegion()
     // call.
     // Otherwise, we have to have the call.
     if (defaultRegion == null) {
       try {
         defaultRegion = new com.amazonaws.regions.DefaultAwsRegionProviderChain().getRegion();
         if (defaultRegion == null) {
-          throw new AwsCryptoException("Default Region was not set and Default AWS Region Provider Chain could not load one.");
+          throw new AwsCryptoException(
+              "Default Region was not set and Default AWS Region Provider Chain could not load one.");
         }
       } catch (SdkClientException ex) {
-        throw new AwsCryptoException("Default Region was not set and Default AWS Region Provider Chain could not load one.", ex);
+        throw new AwsCryptoException(
+            "Default Region was not set and Default AWS Region Provider Chain could not load one.",
+            ex);
       }
     }
     this.defaultRegion_ = defaultRegion;
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
@@ -192,14 +192,8 @@ public final class AwsKmsMrkAwareMasterKeyProvider
      * @see KmsMasterKeyProvider.Builder#buildDiscovery()
      */
     public AwsKmsMrkAwareMasterKeyProvider buildDiscovery() {
-      String sdkDefaultRegion = getSdkDefaultRegion();
-
       if (defaultRegion_ == null) {
-        defaultRegion_ = sdkDefaultRegion;
-      }
-
-      if (discoveryMrkRegion_ == null) {
-        discoveryMrkRegion_ = sdkDefaultRegion;
+        defaultRegion_ = getSdkDefaultRegion();
       }
 
       final boolean isDiscovery = true;
@@ -251,14 +245,8 @@ public final class AwsKmsMrkAwareMasterKeyProvider
      * @see KmsMasterKeyProvider.Builder#buildStrict(List)
      */
     public AwsKmsMrkAwareMasterKeyProvider buildStrict(List<String> keyIds) {
-      String sdkDefaultRegion = getSdkDefaultRegion();
-
       if (defaultRegion_ == null) {
-        defaultRegion_ = sdkDefaultRegion;
-      }
-
-      if (discoveryMrkRegion_ == null) {
-        discoveryMrkRegion_ = sdkDefaultRegion;
+        defaultRegion_ = getSdkDefaultRegion();
       }
 
       final boolean isDiscovery = false;
--- a/src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
@@ -49,12 +49,12 @@ public final class AwsKmsMrkAwareMasterKeyProvider
   private final String defaultRegion_;
 
   public static class Builder implements Cloneable {
-    private String defaultRegion_ = getSdkDefaultRegion();
+    private String defaultRegion_ = null;
     private Optional<KmsMasterKeyProvider.RegionalClientSupplier> regionalClientSupplier_ =
         Optional.empty();
     private AWSKMSClientBuilder templateBuilder_ = null;
     private DiscoveryFilter discoveryFilter_ = null;
-    private String discoveryMrkRegion_ = this.defaultRegion_;
+    private String discoveryMrkRegion_ = null;
 
     Builder() {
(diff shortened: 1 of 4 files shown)


src/test/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProviderTest.java
```diff
@@ -225,33 +224,23 @@ public void no_duplicate_mrks() {
...
+          // it will pick one up from the environment
+          .withDefaultRegion(null)
           .buildStrict("mrk-edb7fe6942894d32ac46dbb1c922d574");
     }
 
     @Test
-    // = compliance/framework/aws-kms/aws-kms-mrk-aware-master-key-provider.txt#2.6
-    // = type=test
-    // # If an AWS SDK Default Region can not be
-    // # obtained initialization MUST fail.
```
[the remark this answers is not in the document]
imabhichow: The check is present because `KmsMasterKeyProvider` allowed `_defaultregion` to be set to null during the build. This is no longer the case with this change. By deferring the call to `getSdkDefaultRegion()` unless no default region is provided, `_defaultregion` will always be a non-null value unless `DefaultAwsCredentialsProvider` could not fetch the region.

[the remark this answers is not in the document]
texastony: Wait... 
```
    private static String getSdkDefaultRegion() {
      try {
        return new com.amazonaws.regions.DefaultAwsRegionProviderChain().getRegion();
      } catch (SdkClientException ex) {
        return null;
      }
    }
```
This method literally returns `null` in case of an Exception. 
I think this introduces an NPE...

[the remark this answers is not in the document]
texastony: No, b/c it was already set to be this way.
Darwin has just moved the call further down the process, as the requester has asked.

texastony (dismissed):

texastony:
@seebees @imabhichow I refactored this to maintain the previous behavior w.r.t. ensuring a region is set by the user IF they pass any KeyIDs without a Region.

@imabhichow your solution broke that behavior,
hence you had to change the unit tests. 

However, the best solution maybe to use the MPL.
We should audit the MPL to see when a Default a Region is loaded.
If the MPL's solution is already optimal,
we can direct the requester of #1768 to use the MPL instead.

This would mean they would not have to wait for a new minor release of the ESDK-Java.

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -413,6 +423,25 @@ private AwsKmsMrkAwareMasterKeyProvider(
     }
 
     this.regionalClientSupplier_ = supplier;
+    // NPE would be thrown via result of extractRegion passed to KMS Client request around line 583.
```
seebees, line 426: /NPE/Nitro Enclave/s right?

src/test/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProviderTest.java
```diff
@@ -224,23 +225,32 @@ public void no_duplicate_mrks() {
...
+          AwsCryptoException.class,
+          () ->
+              AwsKmsMrkAwareMasterKeyProvider.builder()
+                  .withDefaultRegion(null)
+                  .buildStrict("mrk-edb7fe6942894d32ac46dbb1c922d574"));
       AwsKmsMrkAwareMasterKeyProvider.builder()
-          // If default region is set to `null` or not configured
-          // it will pick one up from the environment
-          .withDefaultRegion(null)
+          .withDefaultRegion("us-east-1")
```
seebees, line 235: I don't understand this test. The name is "always need a region", but a region is passed, so I'm confused.

seebees: Wait! Is this saying "See, `null` fails and `us-east-1` passes, must be right?

This is putting back the test right?

seebees (commented):
If my comment is right, I think this is GTG

seebees:
Also, note: this is not a problem in Keyrings.
https://github.com/awslabs/aws-encryption-sdk-specification/blob/a10e0d8889618853f6960ee5f9d622c6d43bc15e/framework/aws-kms/aws-kms-multi-keyrings.md?plain=1#L77-L79

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
texastony: Null Pointer Exception

src/test/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProviderTest.java
texastony: That is correct.
Because this library has a weird feature where `mrk-edb7fe6942894d32ac46dbb1c922d574` is 
considered a valid ARN if and only if a region is given.

seebees (approved):
LGTM

src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -193,6 +193,25 @@ public AwsKmsMrkAwareMasterKeyProvider.Builder withClientBuilder(AWSKMSClientBui
...
+        // default region is found. Because a default region _may_ not be needed.
+        // Note: Final Null check is at
+        // com.amazonaws.encryptionsdk.kms.AwsKmsMrkAwareMasterKeyProvider#AwsKmsMrkAwareMasterKeyProvider#L417
+        try {
+          // = compliance/framework/aws-kms/aws-kms-mrk-aware-master-key-provider.txt#2.6
+          // # In
+          // # discovery mode if a default MRK Region is not configured the AWS SDK
+          // # Default Region MUST be used.
+          discoveryMrkRegion_ =
+              new com.amazonaws.regions.DefaultAwsRegionProviderChain().getRegion();
```
imabhichow, line 208: Since we're directly calling Default Provider Chain, we could remove `getSdkDefaultRegion`
```
private static String getSdkDefaultRegion() {
      try {
        return new com.amazonaws.regions.DefaultAwsRegionProviderChain().getRegion();
      } catch (SdkClientException ex) {
        return null;
      }
    }
```

src/test/java/com/amazonaws/encryptionsdk/kmssdkv2/AwsKmsMrkAwareMasterKeyProviderTest.java
```diff
@@ -256,7 +255,7 @@ public void always_need_a_region() {
     // # obtained initialization MUST fail.
     public void discovery_region_can_not_be_null() {
       assertThrows(
-          IllegalArgumentException.class,
+          AwsCryptoException.class,
```
imabhichow, line 258: This test would never throw an exception, even though we set `discoveryMrkRegion` set to `null` in builder. It will try to fetch Default region from Credentials Provider chain with the new changes.

imabhichow (commented):
We could also consider these two scenarios.

1. What is the expected behaviour if we don't set a `defaultRegion` in builder? Currently, we would try to resolve to the SDK Default Region. With the new changes, we would no longer attempt to resolve the region, and the `defaultRegion` remains `null`. 
 This is the reason for other test failure
```
AwsKmsMrkAwareMasterKeyProviderTest$getMasterKey.other_basic_uses:436 ? AwsCrypto Can't use non-ARN key identifiers or aliases when no default region is set
```
2. What is the expected behaviour when we set `defaultRegion` to `null` in builder? Currently, `defaultRegion` would remain `null`. With new changes, the behaviour would remain the same.

src/test/java/com/amazonaws/encryptionsdk/kmssdkv2/AwsKmsMrkAwareMasterKeyProviderTest.java
texastony: Oh, your right, I went too far.
We should revert this:
https://github.com/aws/aws-encryption-sdk-java/blob/69cfee6cfc9403335925a35fd35717aecce94c7d/src/main/java/com/amazonaws/encryptionsdk/kmssdkv2/AwsKmsMrkAwareMasterKeyProvider.java#L170-L188

texastony (changes requested):
Blocking Merge until this is resolved:
https://github.com/aws/aws-encryption-sdk-java/pull/1964/files#r1515103849

texastony:
> 1. With the new changes, we would no longer attempt to resolve the region, and the `defaultRegion` remains `null`.

I don't believe this correct. I think all end cases set `defaultRegion`, just post the builder, in the `AwsKmsMrkAwareMasterKeyProvider` method.

src/main/java/com/amazonaws/encryptionsdk/kmssdkv2/AwsKmsMrkAwareMasterKeyProvider.java
```diff
@@ -348,6 +352,26 @@ private AwsKmsMrkAwareMasterKeyProvider(
...
+        if (defaultRegion == null) {
+          throw new AwsCryptoException(
+              "Default Region was not set and Default AWS Region Provider Chain could not load one.");
+        }
+      } catch (SdkClientException ex) {
+        throw new AwsCryptoException(
+            "Default Region was not set and Default AWS Region Provider Chain could not load one.",
+            ex);
+      }
+    }
```
imabhichow, line 374: I think we should move this behaviour before we check for non ARN's KeyIds either in Builder or before the check. Otherwise, we would still expect exception being thrown.

texastony: I do not follow. 
Doesn't this check have to happen AFTER the non ARN Key IDs?
Or the illegal combination of "no region with non ARN" cannot be detected.

I believe the intention is to prevent users from providing non ARN Key IDs if the region is not set.
That is what the Unit tests assert.

imabhichow: I guess you're correct. 

However, there are [other unit tests](https://github.com/aws/aws-encryption-sdk-java/blob/e94ee8538501f270138feab0a6138bed7ce640ea/src/test/java/com/amazonaws/encryptionsdk/kmssdkv2/AwsKmsMrkAwareMasterKeyProviderTest.java#L489-L507) that check if the default Region is not configured and are passing non-ARN keys. It should resolve with the AWS SDK region. I believe the original changes were asserted such that [defaultRegion == AWS SDK region](https://github.com/aws/aws-encryption-sdk-java/blob/e94ee8538501f270138feab0a6138bed7ce640ea/src/main/java/com/amazonaws/encryptionsdk/kms/AwsKmsMrkAwareMasterKeyProvider.java#L83). As a result, if defaultMrkRegion is not configured, it will resolve to defaultRegion, and if the AWS SDK Region cannot be fetched, we expect a failure.

texastony: I think this code has evolved to be convoluted.
I do not know what the purpose of these conflicting Unit Tests are trying to yield...

I agree, that the current deployed code behaves as you described it:
"An inferred Region is sufficient to allow for non ARNs", 
at least for STRICT, which is what `other_basic_uses` tests.

I am not sure what the behavior is supposed to be for Discovery,
but I... I guess it is the same: "inferred is sufficient if not provided".

So I was entirely wrong about "no region with non ARN" being an illegal combination.

If we set the region in the Builder, we will need to refactor the unit tests to mock the EC2 data loader to confirm that a null returned there fails the build.

PR #2048, opened 2024-10-06
chore(JavaDocs): detail thread saftey of SDK Client Builders

texastony:
*Issue #, if available:*

The AWS SDK for Java's Service Client builders are not thread safe.
Crypto Tools customer's have reported odd,
occasionally disruptive behavior when using them 
in a concurrent setting.

As such, 
Crypto Tools should make it clear how to set a KMS Client
in a thread safe manner.

Which can be done via the Regional Client Supplier.

Additionally, 
we can improve on the Java Documentation to highlight that
Builders are not thread safe,
and that proper Builder suppliers MUST be used,
or customers SHOULD use the alternative client configuration methods.

*Description of changes:*

# Check any applicable:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,14 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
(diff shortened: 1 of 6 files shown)


seebees (approved):
LGTM

Issue #2112, opened 2025-05-27
Encryption output gets truncated because of incorrect offset handling by FrameEncryptionHandler


alex122830:
### Problem:

The method `FrameEncryptionHandler::processBytes` behaves incorrectly when its parameter `off` is not zero. It leads to the truncated encryption.
This is a unit test that reproduces the issue. Notice that if `startOffset` was `0`, then the test would pass.

```java
    @Test
    public void testStream() throws Exception {
        var testDataString = "Hello, World! ".repeat(10_000);

        int startOffset = 100; // The data will start from this offset
        byte[] inputData = new byte[10_000];
        System.arraycopy(
                testDataString.getBytes(StandardCharsets.UTF_8),
                0,
                inputData,
                startOffset,
                inputData.length - startOffset);
        int originalLength = inputData.length - startOffset;

        // Generate a random AES key
        var rnd = new SecureRandom();
        byte[] rawKey = new byte[16];
        rnd.nextBytes(rawKey);
        var cryptoKey = new SecretKeySpec(rawKey, "AES");

        // Create a keyring using the generated AES key
        var materialProviders = MaterialProviders.builder()
                .MaterialProvidersConfig(MaterialProvidersConfig.builder().build())
                .build();
        var keyringInput = CreateRawAesKeyringInput.builder()
                .wrappingKey(ByteBuffer.wrap(cryptoKey.getEncoded()))
                .keyNamespace("Example")
                .keyName("RandomKey")
                .wrappingAlg(AesWrappingAlg.ALG_AES128_GCM_IV12_TAG16)
                .build();
        var keyring = materialProviders.CreateRawAesKeyring(keyringInput);
        var crypto = AwsCrypto.standard();

        // Encrypt the data
        byte[] encryptedData;
        try (var os = new ByteArrayOutputStream()) {
            try (var cryptoOutput = crypto.createEncryptingStream(keyring, os, Map.of())) {
                cryptoOutput.write(inputData, startOffset, inputData.length - startOffset);
            }
            encryptedData = os.toByteArray();
        }

        // Decrypt the data
        int decryptedLength = 0;
        byte[] decryptedData = new byte[inputData.length];
        try (var is = new ByteArrayInputStream(encryptedData);
             var cryptoInput = crypto.createDecryptingStream(keyring, is)) {
            int offset = startOffset;
            do {
                int bytesRead = cryptoInput.read(decryptedData, offset, decryptedData.length - offset);
[...]

PR #2113, opened 2025-05-27
fix: do not truncate encrypted streams when offset is greater than zero

kessplas:
*Issue #, if available:* https://github.com/aws/aws-encryption-sdk-java/issues/2112

*Description of changes:* Fixes the truncation bug described in the issue above. I think using parens and subtracting offset is easier to read than the suggested change but is logically the same. We could also introduce a new variable e.g. `bytesProcessed` which starts at 0 and counts up by `size` each loop, but I think this approach is most readable. 

# Check any applicable:

### DIFF
--- a/src/test/java/com/amazonaws/encryptionsdk/internal/FrameEncryptionHandlerTest.java
+++ b/src/test/java/com/amazonaws/encryptionsdk/internal/FrameEncryptionHandlerTest.java
@@ -165,7 +165,7 @@ public class FrameEncryptionHandlerTest {
     int startOffset = 100; // The data will start from this offset
     byte[] inputDataWithOffset = new byte[10_000];
     // the length of the actual data
-    int dataLength = dataLength;
+    int dataLength = inputDataWithOffset.length - startOffset;
     // copy some data, starting at the startOffset
     // so the first |startOffset| bytes are 0s
     System.arraycopy(


src/main/java/com/amazonaws/encryptionsdk/internal/FrameEncryptionHandler.java
```diff
@@ -116,7 +116,9 @@ public ProcessingSummary processBytes(
       // update offset by the size of bytes being encrypted.
       offset += size;
       // update size to the remaining bytes starting at offset.
-      size = len - offset;
+      // the original offset MUST NOT be part of the length calculation,
+      // or else the encryption will be truncated.
+      size = len - (offset - off);
```
seebees, line 121:
```suggestion
      // `offset` is a local variable that accumulates data the encrypted data,
      // and it is initialized with `off`.
      // Therefore the original offset, `off`, MUST be removed
      // or else the encryption will be truncated.
      size = (len - offset) - off;
```

src/test/java/com/amazonaws/encryptionsdk/internal/FrameEncryptionHandlerTest.java
```diff
@@ -117,4 +132,87 @@ private void assertHeaderNonce(byte[] expectedNonce, byte[] buf) {
...
   }
+
+  /**
+   * This isn't a unit test, but it reproduces a bug in the FrameEncryptionHandler where the stream
+   * would be truncated when the offset is >0
+   *
+   * @throws Exception
+   */
+  @Test
+  public void testStreamTruncation() throws Exception {
```
seebees, line 143: The test could use some more comments to explain how you are verifying what is going on.

seebees (changes requested):
Left a suggestion, I would like to add a few comments to the tests.
We could do that in a different PR if you are in a hurry though

src/test/java/com/amazonaws/encryptionsdk/internal/FrameEncryptionHandlerTest.java
```diff
@@ -117,4 +132,87 @@ private void assertHeaderNonce(byte[] expectedNonce, byte[] buf) {
...
+            .build();
+    IKeyring keyring = materialProviders.CreateRawAesKeyring(keyringInput);
+    AwsCrypto crypto = AwsCrypto.standard();
+
+    // Encrypt the data
+    byte[] encryptedData;
+    try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+      try (CryptoOutputStream cryptoOutput =
+          crypto.createEncryptingStream(keyring, os, Collections.emptyMap())) {
+        cryptoOutput.write(inputData, startOffset, inputData.length - startOffset);
```
seebees, line 190:
```suggestion
        // Here we use the offset to encrypt only part of the inputData
        cryptoOutput.write(inputData, startOffset, inputData.length - startOffset);
```

src/test/java/com/amazonaws/encryptionsdk/internal/FrameEncryptionHandlerTest.java
```diff
@@ -117,4 +132,87 @@ private void assertHeaderNonce(byte[] expectedNonce, byte[] buf) {
...
+    assertArrayEquals(expectedOutput, nonStreamDecrypt.getResult());
+
+    // Check streaming decrypt
+    int decryptedLength = 0;
+    byte[] decryptedData = new byte[inputData.length];
+    try (ByteArrayInputStream is = new ByteArrayInputStream(encryptedData);
+        CryptoInputStream cryptoInput = crypto.createDecryptingStream(keyring, is)) {
+      int offset = startOffset;
+      do {
+        int bytesRead = cryptoInput.read(decryptedData, offset, decryptedData.length - offset);
```
seebees, line 207: I would expect that this offset would mean I don't decrypt all the data? I'm not sure I understand this.

src/main/java/com/amazonaws/encryptionsdk/internal/FrameEncryptionHandler.java
seebees: `offset` ==> `position`

seebees (approved):
LGTM

seebees (approved):

PR #2174, opened 2025-12-17
perf: avoid expensive call to Mac.getInstance

mkeskells:
*Issue #, if available:*
No Issue

*Description of changes:*
Minimal change to avoid calls to Mac.getInstance
Call to this were only used to extract the provider, which I though would be better expressed as an enum, as there are very few MacAlgorithms in use.
In my (closed source) benchmark this call accounts for 5.6% of the CPU time taken for a decryption (a call to `AwsCrypto.decryptData`), with small payloads, and thee same for encryption

I cant fully build this locally so cant run a test cycle prior to submitting this PR

# Check any applicable:

### DIFF
--- a/src/main/java/com/amazonaws/encryptionsdk/CryptoAlgorithm.java
+++ b/src/main/java/com/amazonaws/encryptionsdk/CryptoAlgorithm.java
@@ -7,6 +7,7 @@ import com.amazonaws.encryptionsdk.exception.BadCiphertextException;
 import com.amazonaws.encryptionsdk.internal.CommittedKey;
 import com.amazonaws.encryptionsdk.internal.Constants;
 import com.amazonaws.encryptionsdk.internal.HmacKeyDerivationFunction;
+import com.amazonaws.encryptionsdk.internal.MacAlgorithm;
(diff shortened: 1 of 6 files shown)


mkeskells:
I haven't changed the javadoc to match the implementation, as it seems that a first review is likely to have an opinion as to API changes

mkeskells:
Hi
Are any reviewers/maintainers available to make comment or progress this PR?

ShubhamChaturvedi7:
I've approved the CI to run but looks like this is a breaking change. Let's wait for the CI results.

mkeskells:
Thanks
I will have a look this evening

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

mkeskells:
@ShubhamChaturvedi7 thanks. The CI failures loom to me to be unrelated to my change - failing in `Configure AWS Credentials`

I had hoped that this would not be a breaking change. It only changes non public fields and internal classes, but I dont know the threshold for breaking changes for this library!

Happy to work with you or someone else to lessen the breaks which reducing the CPU cost, 

Please advise of the best next steps

